PR #8, opened 2021-05-25
docs: add Apache-2.0 license

jaredallard:
**What this PR does / why we need it**: As it says, it does!

**JIRA ID**: DT-0

**Notes for your reviewer**:


george-e-shaw-iv (approved):
you don't need a license to drive a sandwich

PR #29, opened 2021-06-22
feat: remove docker box :latest images on provision

jaredallard:
**What this PR does / why we need it**: This PR adds automatic removal of `:latest` docker images on the configured `box` image registry on `devenv provision`. This was in the OG devenv as users seemed to expect this behaviour, it was slated to be added back post-MVP. So, this adds it back (now in Go!)

**JIRA ID**: DT-0

**Notes for your reviewer**:


deregtd (approved):
RIP my internet

PR #33, opened 2021-06-30
fix: remove TTY argument in remove images

jaredallard:
**What this PR does / why we need it**: This PR removes the `-it` (interactive, tty) request to the `docker exec` call for removing docker images. This would sometimes trip up and cause the command to fail. 

**JIRA ID**: DT-0

**Notes for your reviewer**: [**Context**](https://outreach-hq.slack.com/archives/CN9MU7GLW/p1625079167475400)

### DIFF
--- a/pkg/embed/config/kind.yaml
+++ b/pkg/embed/config/kind.yaml
@@ -3,7 +3,7 @@ apiVersion: kind.x-k8s.io/v1alpha4
 name: "{{ .Name }}"
 nodes:
   - role: control-plane
-    image: "gcr.io/outreach-docker/kindest/node:v1.20.5{{ .TagSuffix }}"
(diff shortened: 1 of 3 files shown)


pkg/embed/config/kind.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: kind.x-k8s.io/v1alpha4
 name: "{{ .Name }}"
 nodes:
   - role: control-plane
-    image: "gcr.io/outreach-docker/kindest/node:v1.20.5{{ .TagSuffix }}"
+    image: "gcr.io/outreach-docker/kindest/node:v1.20.7"
```
jaredallard, line 6: This is needed for the arm64 support, tested it on x86 and arm64 w/ no issues so it should be fine :shipit:

deregtd (approved):

PR #52, opened 2021-08-10
feat: localizer enhancements [DTSS-595]

george-e-shaw-iv:
**What this PR does / why we need it**: 
* Brings in all of the localizer enhancements that were brought about via DTSS-595.

**JIRA ID**: [DTSS-595]

**Notes for your reviewer**:

[DTSS-595]: https://outreach-io.atlassian.net/browse/DTSS-595


pkg/kubernetestunnelruntime/localizer.go
```diff
@@ -15,8 +14,6 @@ var (
 	LocalizerDownloadURL = "https://github.com/jaredallard/localizer/releases/download/" +
```
deregtd, line 14: should we repoint this to getoutreach?

deregtd (approved):

PR #56, opened 2021-08-17
feat: bump localizer version

george-e-shaw-iv:
**What this PR does / why we need it**: 
* turns out you need to update the go.mod version and this version. whoops

**JIRA ID**: XX-XX

**Notes for your reviewer**:


jaredallard (approved):
lolz, we could technically read this from the go.mod if we want? since we can use go module information on runtime

PR #103, opened 2021-10-05
fix(destroy): configure kubernetes runtime on destroy

jaredallard:

**JIRA ID**: XX-XX


george-e-shaw-iv (approved):
kooberneties

PR #123, opened 2021-10-26
fix(local-app): redeploy applications that o.CreateManifests

jaredallard:

Title says, it does. Also we now skip Shutdown/OutOfcpu/OutOfpods pods in the wait for ready check, and the monitoring namespace.

**JIRA ID**: DTSS-0


pkg/devenvutil/devenv.go
```diff
@@ -180,11 +180,25 @@ func FindUnreadyPods(ctx context.Context, k kubernetes.Interface) ([]string, err
...
 		if po.Status.Phase == corev1.PodSucceeded {
 			continue
 		}
 
+		// handle cloud edge cases
+		if po.Status.Reason == "OutOfcpu" ||
+			po.Status.Reason == "OutOfpods" ||
+			po.Status.Reason == "Shutdown" {
+			continue
+		}
```
george-e-shaw-iv, line 200: should we at least warn for transparency/visibility?

george-e-shaw-iv (approved):

PR #127, opened 2021-11-03
fix: actually use Vault address from box

jaredallard:

This PR fixes the devenv to actually use the Vault addr from box which it silently wasn't using. This also moves us off of the official Vault SDK and instead onto `vault-client` for ease of use.

**JIRA ID**: DTSS-0


cmd/devenv/provision/prepare.go
```diff
@@ -184,12 +180,12 @@ func (o *Options) ensureImagePull(ctx context.Context) error {
 		return errors.Wrap(err, "failed to create vault client")
 	}
 
-	sec, err := v.Logical().Read(secretPath)
+	paths := strings.Split(o.b.DeveloperEnvironmentConfig.ImagePullSecret, "/")
+	sec, err := v.GetKV2Secret(ctx, paths[0], strings.Join(paths[1:], "/"))
```
george-e-shaw-iv, line 184: can you guarantee this will never panic?

jaredallard: Yep

george-e-shaw-iv (approved):

PR #128, opened 2021-11-03
feat: stop/start remote clusters

jaredallard:

This PR adds support for `devenv stop` and `devenv start` to remote Kubernetes clusters, we also now cleanup loft spaces to prevent extra resources from being left over when a cluster is deleted.

**JIRA ID**: DTSS-0


george-e-shaw-iv (approved):
much cleaner with these interfaces too, nice

PR #129, opened 2021-11-04
fix(context): show kind always except unprov

jaredallard:

This PR fixes `devenv context` to show KinD always, except when it's unprovisioned to ensure that there is a decent UX for cases when it's stopped but undetectable.

**JIRA ID**: [DTSS-0]

### DIFF
--- a/pkg/kubernetesruntime/kind.go
+++ b/pkg/kubernetesruntime/kind.go
@@ -219,8 +219,8 @@ func (kr *KindRuntime) GetKubeConfig(ctx context.Context) (*api.Config, error) {
 func (kr *KindRuntime) GetClusters(ctx context.Context) ([]*RuntimeCluster, error) {
 	curStatus := kr.Status(ctx).Status.Status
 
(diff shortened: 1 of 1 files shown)


pkg/kubernetesruntime/kind.go
```diff
@@ -219,8 +219,8 @@ func (kr *KindRuntime) GetKubeConfig(ctx context.Context) (*api.Config, error) {
 func (kr *KindRuntime) GetClusters(ctx context.Context) ([]*RuntimeCluster, error) {
 	curStatus := kr.Status(ctx).Status.Status
 
-	if curStatus == status.Unprovisioned || curStatus == status.Unknown {
-		// Only return a cluster if it's actively running
+	if curStatus == status.Unprovisioned {
+		// Only return a cluster if it's no unprovisioned
```
jaredallard, line 223:
```suggestion
		// Only return a cluster if it's not unprovisioned
```

PR #133, opened 2021-11-11
feat: new registry get command get preferred image registry, support multiple loft clusters

jaredallard:

This PR adds a `devenv registry get` command meant to be used by Tilt (and maybe other tooling?) to get the preferred address/path for a development docker image registry using the new `gobox/pkg/region` library. This is then also used to lay the framework for supporting that in `loft` (not actually hooked up to our new multi-region cluster)

**JIRA ID**: XX-XX


vk-outreach:
lgtm from golang p.o.v

george-e-shaw-iv (approved):

PR #135, opened 2021-11-12
feat(helm): upgrade to latest to enable resizable local-path volume

nissim-natanov-outreach:
rancher enabled resizable volume flag on its storage few months ago, hope latest image brings this change in.

**JIRA ID**: QSS-00


nissim-natanov-outreach:
new rancher helm controller does not set the flag ... I might need to patch it in directly

jaredallard:
Offline decided not to do this

PR #137, opened 2021-11-12
fix(snapshot): dont try to snapshot ourself (minio)

jaredallard:

This PR does what the title says. We basically started to snapshot our self when we moved minio into the container randomly increasing the size of the snapshots. Pretty crazy.

**JIRA ID**: [DTSS-0]

### DIFF
--- a/cmd/devenv/snapshot/snapshot.go
+++ b/cmd/devenv/snapshot/snapshot.go
@@ -447,7 +447,7 @@ func (o *Options) CreateSnapshot(ctx context.Context) (string, error) { //nolint
 			Name: backupName,
 		},
 		Spec: velerov1api.BackupSpec{
-			// Don't include velero, we need to install it before the backup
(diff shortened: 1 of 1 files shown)


cmd/devenv/snapshot/snapshot.go
```diff
@@ -448,7 +448,7 @@ func (o *Options) CreateSnapshot(ctx context.Context) (string, error) { //nolint
 		},
 		Spec: velerov1api.BackupSpec{
 			// Don't include velero, we need to install it before the backup
```
berickson1, line 450: Nit: consider updating this comment

jaredallard, line 450:
```suggestion
			// Don't include velero, we need to install it before the backup. Skip minio because it's the snapshot backend
```

berickson1 (approved):

PR #141, opened 2021-11-19
fix: Enable devenv completion to work without updating.

ChrisPRobinson:
[DTSS-429]

fix: Enable devenv completion to work without updating.

**JIRA ID**:DTSS-429

## Notes for your reviewers

[DTSS-429]: https://outreach-io.atlassian.net/browse/DTSS-429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/devenv/devenv.go
+++ b/cmd/devenv/devenv.go
@@ -73,7 +73,7 @@ func main() {
 			// in before to see if that command is being run, you would be wrong
 			// Using the args passed in to see if the completion command
 			// was provided. Other global flags are just ignored.
-			if c.Args().Len() > 0 && c.Args().First() == "completion" {
+			if c.Args().First() == "completion" {
 				return nil
 			}
 
--- a/cmd/devenv/devenv.go
+++ b/cmd/devenv/devenv.go
@@ -69,6 +69,14 @@ func main() {
 		`),
 		EnableBashCompletion: true,
 		Before: func(c *cli.Context) error {
+			// NOTE: You would think you can check the the c.Command.Name == "completion"
+			// in before to see if that command is being run, you would be wrong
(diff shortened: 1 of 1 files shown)


cmd/devenv/devenv.go
```diff
@@ -69,6 +69,14 @@ func main() {
 		`),
 		EnableBashCompletion: true,
 		Before: func(c *cli.Context) error {
+			// NOTE: You would think you can check the the c.Command.Name == "completion"
```
coding-paras, line 72: Wow. I wonder why does this doesn't work.

cmd/devenv/devenv.go
```diff
@@ -69,6 +69,14 @@ func main() {
...
+			// in before to see if that command is being run, you would be wrong
+			// Using the args passed in to see if the completion command
+			// was provided. Other global flags are just ignored.
+			if c.Args().Len() > 0 && c.Args().First() == "completion" {
```
coding-paras, line 76: [nit] [c.Args().First()](https://github.com/urfave/cli/blob/master/args.go#L28) is nil safe i.e. you don't need to explicitly check `c.Args().Len() > 0`

coding-paras (approved):

george-e-shaw-iv (approved):
approved, but paras is right

cmd/devenv/devenv.go
ChrisPRobinson: Ah, should have looked deeper that the impl.

ChrisPRobinson:
Updated to fix what @coding-paras pointed out.

PR #143, opened 2021-11-25
fix: Using different Minio helm chart

deregtd:

**JIRA ID**: XX-XX


matteddy (approved):
LGTM. Not sure exactly what the version change does but seems like it can't break it worse than status quo

PR #151, opened 2021-12-22
feat: alert when waiting for some time

george-e-shaw-iv:
* alerts when waiting for sudo prompt or aws login after some time

**JIRA ID**: XX-XX

## Notes for your reviewers
* bad branch name, sry

### DIFF
--- a/cmd/devenv/tunnel/tunnel.go
+++ b/cmd/devenv/tunnel/tunnel.go
@@ -99,9 +99,9 @@ func (o *Options) Run(ctx context.Context) error { //nolint:funlen
 			case <-ctx.Done():
 			case <-done:
 				return
-			default:
+			case <-time.After(time.Second * 15):
 				// Sleep for 15 seconds between each alert after the first one.
-				async.Sleep(ctx, time.Second*15)
+				continue
 			}
 		}
 	}(ctx)
--- a/cmd/devenv/tunnel/tunnel.go
+++ b/cmd/devenv/tunnel/tunnel.go
@@ -87,20 +87,21 @@ func (o *Options) Run(ctx context.Context) error { //nolint:funlen
 		return err2
 	}
 
-	var elevatedPermissions bool
+	var done chan struct{}
 	go func(ctx context.Context) {
 		// Sleep for 7 seconds the first time before checking to alert for permissions.
 		async.Sleep(ctx, time.Second*7)
-		for {
-			if elevatedPermissions {
-				// Got elevated permissions, don't alert.
-				return
-			}
 
+		for ctx.Err() == nil {
 			alert.Alert("Your tunnel needs elevated permissions!")
 
-			// Sleep for 15 seconds between each alert after the first one.
-			async.Sleep(ctx, time.Second*15)
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				// Sleep for 15 seconds between each alert after the first one.
+				async.Sleep(ctx, time.Second*15)
+			}
 		}
 	}(ctx)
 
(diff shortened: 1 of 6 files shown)


cmd/devenv/tunnel/tunnel.go
```diff
@@ -86,12 +87,30 @@ func (o *Options) Run(ctx context.Context) error { //nolint:funlen
...
+				return
+			}
+
+			alert.Alert("Your tunnel needs elevated permissions!")
+
+			// Sleep for 15 seconds between each alert after the first one.
+			async.Sleep(ctx, time.Second*15)
+		}
+	}(ctx)
+
```
vk-outreach, line 106: This is not thread safe.  I doubt it matters for this specific tool but the thread safe version is like so (also added a ctx.Err() == nil check):


```suggestion
	var elevatedPermissions int32
	go func(ctx context.Context) {
		// Sleep for 7 seconds the first time before checking to alert for permissions.
		async.Sleep(ctx, time.Second*7)
		for atomic.LoadInt32(&elevatedPermissions) == 0 && ctx.Err() == nil {
			alert.Alert("Your tunnel needs elevated permissions!")

			// Sleep for 15 seconds between each alert after the first one.
			async.Sleep(ctx, time.Second*15)
		}
	}(ctx)

```

vk-outreach: You'd have to replace the `elevatedPermissions = true` with `atomic.AddInt32(&elevatedPermissions, 1)`

vk-outreach: Actually, this is a clear use of signaling, so an alternate approach without using atomic would be via using a channel for signaling:

```golang
var done chan struct{}
go func(ctx context.Context) {
    async.Sleep(time.Second*15)
    for ctx.Err() != nil {
        alert.Alert("...")
        select {
        case <- ctx.Done():  return
        case <- time.After(time.Seconds*15):
        }
    }
}(ctx)

...
close(done)
```

george-e-shaw-iv: done

george-e-shaw-iv:
lol i put a bug in the code one second

george-e-shaw-iv:
k now the bug is gone

cmd/devenv/tunnel/tunnel.go
```diff
@@ -86,12 +87,32 @@ func (o *Options) Run(ctx context.Context) error { //nolint:funlen
...
+		for ctx.Err() == nil {
+			alert.Alert("Your tunnel needs elevated permissions!")
+
+			select {
+			case <-ctx.Done():
+			case <-done:
+				return
+			default:
+				// Sleep for 15 seconds between each alert after the first one.
+				async.Sleep(ctx, time.Second*15)
```
vk-outreach, line 104:
```suggestion
			case <- time.After(time.Second*15):
```

george-e-shaw-iv: Is it not better to use the async function so we can respect context cancellations immediately?

vk-outreach (approved):
LGTM

cmd/devenv/tunnel/tunnel.go
george-e-shaw-iv: wait im a moron don't respond

vk-outreach: The cancellation will trigger the other `<- ctx.Done()` clause -- so it will cancel immediately with the way the select is written

george-e-shaw-iv: yes i realized that immediately after commenting 😆

PR #167, opened 2022-01-19
feat: auth-vault reauthenticates vault operator in devenv

plisy:
When vault token in devenv expires, new secrets on deployment don't get created. There's no way to just refresh the token today. `deploy-app` does the trick, but it also applies manifests (in case of cloud devenv, without pushing docker image). This helps mitigate issue with `devspace deploy`.

**JIRA ID**: XX-XX

### DIFF
--- a/cmd/devenv/auth-vault/auth.go
+++ b/cmd/devenv/auth-vault/auth.go
@@ -1,0 +1,96 @@
+package auth
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/getoutreach/devenv/internal/vault"
+	"github.com/getoutreach/devenv/pkg/cmdutil"
+	"github.com/getoutreach/devenv/pkg/config"
+	"github.com/getoutreach/devenv/pkg/devenvutil"
+	"github.com/getoutreach/devenv/pkg/kube"
+	"github.com/getoutreach/gobox/pkg/box"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
+	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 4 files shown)


george-e-shaw-iv (commented):
This is great and I'll leave the final review for Jared, but I'm just wondering for UX and leaving a door open for ourselves in the future if we wanted to rebrand this under something like `devenv auth refresh`. I think we could eventually end up doing other things with `devenv auth`? Not sure tho. I'll leave final opinion to Jared

jaredallard:
> This is great and I'll leave the final review for Jared, but I'm just wondering for UX and leaving a door open for ourselves in the future if we wanted to rebrand this under something like `devenv auth refresh`. I think we could eventually end up doing other things with `devenv auth`? Not sure tho. I'll leave final opinion to Jared

I think agree here, having this be `devenv auth refresh` is probably a good idea.

cmd/devenv/auth-vault/auth.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	return &Options{
+		k:    k,
+		conf: conf,
+		log:  log,
+	}, nil
+}
+
+func NewCmdAuthVault(log logrus.FieldLogger) *cli.Command {
```
jaredallard, line 52: nit: all of these exported functions need a comment, this is going to have lintroller be ran on it very soon

jaredallard (commented):
lgtm from a code standpoint, but dropping the approval until we hash out `auth refresh` path.

cmd/devenv/auth-vault/auth.go
plisy: done

jaredallard (approved):

PR #173, opened 2022-01-25
feat: Added `devenv devspace` proxy command

plisy:
Adding `devspace` command as proxy for using `devspace` with `dev-environment` context.

**JIRA ID**: XX-XX

### DIFF
--- a/cmd/devenv/devspace/devspace.go
+++ b/cmd/devenv/devspace/devspace.go
@@ -1,0 +1,112 @@
+package devspace
+
+import (
+	"context"
+	"os"
+	"os/exec"
+
+	"github.com/getoutreach/devenv/internal/vault"
+	"github.com/getoutreach/devenv/pkg/config"
+	"github.com/getoutreach/devenv/pkg/devenvutil"
+	"github.com/getoutreach/devenv/pkg/kube"
+	"github.com/getoutreach/gobox/pkg/box"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+type Options struct {
+	log  logrus.FieldLogger
+	k    kubernetes.Interface
+	conf *rest.Config
+
+	args []string
+}
+
+func NewOptions(log logrus.FieldLogger) (*Options, error) {
+	k, conf, err := kube.GetKubeClientWithConfig()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create kubernetes client")
+	}
+
+	return &Options{
+		k:    k,
+		conf: conf,
+		log:  log,
+	}, nil
+}
+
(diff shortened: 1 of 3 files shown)


cmd/devenv/devspace/devspace.go
```diff
@@ -0,0 +1,112 @@
...
+
+			return o.Run(c.Context)
+		},
+	}
+}
+
+func (o *Options) Run(ctx context.Context) error {
+	_, err := exec.LookPath("devspace")
+	if err != nil {
+		return errors.Wrap(err, "failed to find devspace CLI. run `orc setup` to install it")
```
jaredallard, line 62: orc isn't open-source, can't mention it here. Use the downloader framework instead like we do w/ kind, loft, and other deps.

cmd/devenv/devspace/devspace.go
```diff
@@ -0,0 +1,112 @@
...
+	if !contextSet {
+		o.args = append(o.args, "--kube-context", "dev-environment")
+	}
+	// We are specifying namespaces in devspace.yaml and kube context here. No need to frighten people with useless warnings.
+	if !noWarn && !contextSet {
+		o.args = append(o.args, "--no-warn")
+	}
+
+	//nolint:gosec // Why: This is a pass through command to devspace, we want to pass through all arguments.
+	cmd := exec.CommandContext(ctx, "devspace", o.args...)
```
jaredallard, line 105: use `cmdutil.RunKubernetesCommand` here instead, that does the same as `--kube-context` but ensure that it only has access to the correct KUBECONFIG

cmd/devenv/devspace/devspace.go
```diff
@@ -0,0 +1,112 @@
...
+
+	cmd.Stdout = os.Stdout
+	cmd.Stdin = os.Stdin
+	cmd.Stderr = os.Stderr
+
+	return cmd.Run()
```
jaredallard, line 111: decorate this error, sometimes it's not clear when it says `failed to run: exit code 1`, use `errors.Wrap(cmd.Run(), "failed to wrap devspace")` or something like that.

jaredallard (changes requested):

jaredallard:
I'm not really sure how I feel about adding a devspace specific command to this when we've specifically avoided being specific to implementations in all of these commands. I'll need to think about this for a bit.

plisy:
The idea was to follow `kubectl` as example. This way we can experiment with using devspace, without having to deal with context issues and integrating it fully into devenv commands.

I plan to integrate it fully, but that's a longer effort. In the meantime we can have people trying it out and reporting issues.

jaredallard:
Yeah, but once we commit to `devenv devspace` as a signature we need to then make a breaking change to remove it later which I don't really want to do any time soon.

Can we rename this to `devenv apps dev` and just stub the `apps` part? I have a larger PR that renamed `deploy-app`, `update-app` and `delete-app` to `apps` counterparts.

Then, in that command, just check if there exists a `devspace.yaml` and run devspace if so. I really don't want to tie this to specific solutions. Hence why `devenv tunnel` is a thing vs `devenv localizer`

PR #176, opened 2022-02-02
fix(apps): detect bootstrap apps via lockfile

jaredallard:

This is better for apps that use `service.yaml` to set dependencies but aren't bootstrapped.

**JIRA ID**: XX-XX

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
   release-dryrun:
     executor:
       name: shared/testbed-machine
-    resource_class: large
+    resource_class: xlarge
     steps:
       - shared/setup_environment:
           machine: true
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -41,7 +41,7 @@ jobs:
   release-dryrun:
     executor:
       name: shared/testbed-machine
-    resource_class: large
+    resource_class: xlarge
```
george-e-shaw-iv, line 44: do you need to mark these as deviations?

george-e-shaw-iv (approved):

.circleci/config.yml
jaredallard: nope, release-dryrun is already deviated

PR #177, opened 2022-02-02
fix: use gobox github pkg for auth

jaredallard:

Title says, it does. This should enable support for different methods of getting a Github token.

**JIRA ID**: DTSS-0


george-e-shaw-iv (approved):

ChrisPRobinson (approved):
LGTM

PR #186, opened 2022-02-07
fix(apps): dont swallow auth error

jaredallard:

Don't swallow 404/auth errors when determining the application type. Originally we swallowed errors to detect the type, but when looking for an invalid repo or on invalid auth this would swallow that error and instead return a cryptic ref failure message.

**JIRA ID**: [DTSS-0]


george-e-shaw-iv (approved):
good stff

PR #187, opened 2022-02-07
fix(apps): full clone repo to prevent flakiness in checking out

jaredallard:

This should fix issues with cloning a partial repo that would sometimes flake when trying to checkout

**JIRA ID**: XX-XX


ChrisPRobinson (approved):
LGTM

PR #192, opened 2022-02-09
feat(deps): update all deps

jaredallard:

Updated all dependencies both `go.mod` and in `embed/pre-restore`.

**JIRA ID**: DTSS-0


Dockerfile.velero-kubectl
```diff
@@ -0,0 +1,32 @@
+# docker buildx build --platform "linux/amd64,linux/arm64" -f Dockerfile.velero-kubectl \
```
jaredallard, line 1: Needed until https://github.com/vmware-tanzu/helm-charts/issues/339 is fixed.

pkg/kubernetesruntime/loft.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	loftVersion     = "v2.0.2"
+	loftVersion     = "v2.1.1"
```
jaredallard, line 31: Confirmed this is working with our older version of loft atm, so we're safe to upgrade

cmd/devenv/provision/prepare.go
```diff
@@ -76,7 +76,7 @@ func (o *Options) deployStage(ctx context.Context, stage string) error { //nolin
 
 	err = devenvutil.WaitForAllPodsToBeReady(ctx, o.k, o.log)
 	if err != nil {
-		return errors.Wrap(err, "failed to wait for pods to be ready w")
```
george-e-shaw-iv, line 79: i think the old error message was better

george-e-shaw-iv (approved):

PR #203, opened 2022-02-24
fix(apps/deploy): correctly resolve legacy application names

jaredallard:

This PR fixes the legacy application name resolving logic to consider the absolute path to the local directory. This was bugged on relative paths, e.g. `.` and now handles them properly.

## Jira ID

[DTSS-0]


edmangimelli-outreach (approved):
good looks me to

PR #210, opened 2022-03-08
docs: Devspace integration proposal

plisy:

Docs for devspace integration design.

## Jira ID

[DTSS-1367]

## Notes for your reviewers

[DTSS-1367]: https://outreach-io.atlassian.net/browse/DTSS-1367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/devspace-proposal/README.md
```diff
@@ -0,0 +1,83 @@
+# Development @ Outreach
+
```
bhogatemadhura, line 2: eventually we do want it the folder to be named `devspace`, is that right ?

The `proposal` makes it sense that it should be a confluence page.

plisy: Usually we put those in google docs, but for some reason I have much less difficulty writing markdown. I can convert it to google doc as well.

bhogatemadhura: lets just have single source of truth .. wherever that is !

docs/devspace-proposal/bootstrap.md
```diff
@@ -0,0 +1,93 @@
...
+## `apps dev`
+
+`devenv` uses `devspace` with the right Kubernetes context and namespace to replace application pods with a dev pod.
+
+> This functionality is the same regardless of type of cluster being used.
+
+1. `devspace` deploys the pre-built app (this ensures the infra, services, ingress, ..., exists, but avoids building the image from source, this saves time)
+1. `devspace` scales down the application deployment based on common labels (image selectors are much more fragile)
+1. `devspace` starts a new pod with same configuration (env variables, secrets, mounts) like the application pods with following changes
+   - based on CI image
```
jaredallard, line 53: We should seperate out the Dockerfile used instead of this. If we make the first stage container a "build container" then we can use that for both CI and devspace. This will allow us to ensure that custom tooling is only in one place needed to build the binary. This does mean we can't "pre-cache" the image, but this is a one time pain with our current plans.

docs/devspace-proposal/README.md
```diff
@@ -0,0 +1,83 @@
...
+Developers can build and start the service, run tests, including e2e.
+
+#### Performance
+
+The development pods are not going to stay around forever and need to be recreated when manifests and Dockerfile change. With cloud devenv they also need to be recreated after the cluster goes to sleep.
+
+This leads to long time to ready/run in such a container. To avoid this, nightly build creates images with latest packages built in, Buildkit and dev pod share package cache, so devs don't have to redownload them all the time.
+
+## Client
```
jaredallard, line 56: Instead of talking about internal services, client, giraffe, flagship, etc. Distill these down into generic application 'profiles'.

docs/devspace-proposal/bootstrap.md
```diff
@@ -0,0 +1,93 @@
...
+   - added `VERSION` env variable
+   - set up credentials from env variables to be used for pulling packages (npm, golang, gems)
+   - command tails a log file (this is where application will `tee` logs so they land in Datadog)
+1. `devspace` syncs local source code into the dev container (and keeps syncing them as long as the session is open)
+1. `devspace` forwards ports (8000 for metrics, 8080 for HTTP, 5000 for gRPC, **TBD** for debugger)
+1. `devspace` opens an interactive terminal into that pod
+
+> Note: make commands understand application is running in dev container and `tee` application logs into the file that is tailed for Datadog logs.
+
+#### `apps dev` source code sync alternative
```
jaredallard, line 69: With my above comment, I don't feel like this path is needed, ultimately.

docs/devspace-proposal/client.md
```diff
@@ -0,0 +1,153 @@
+# Devenv & Client
```
jaredallard, line 1: Again, this is all super internal information that doesn't belong in this repo. Let's create a webapp profile which targets this usecase.

docs/devspace-proposal/client.md
```diff
@@ -0,0 +1,153 @@
...
+| file-loader                 | 2.25 secs           | 0.969 secs         | 0.72 secs          |
+| css-loader                  | 0.413 secs          | 0.155 secs         | 0.13 secs          |
+| html-webpack-plugin         | 0.164 secs          | 0.089 secs         | 0.06 secs          |
+| style-loader and css-loader | 0.067 secs          | 0.033 secs         | 0.018 secs         |
+
+Given these numbers, developing clients in regular dev containers is not feasible. The options are 
+
+* provide special dev containers running on faster nodes (not feasible due to cost, the machines need lots of cores and large premium storage to be fast)
+* reverse port forwarding into cluster and then through ingress to local box again (round trip slows things down on request, port forwarding may be unstable)
+* reverse proxy built into `devenv` decides where to route traffic (cluster vs localhost, and possibly `web-master.outreach-staging.com` so we can skip the whole deployment to dev-environment altogether)
```
jaredallard, line 30: In the short term the second option is likely preferable, "reverse proxy built into devenv" will need to listen on privilege ports to get a 1:1 DNS support. At the moment we round trip through cloud and don't have many issues, so I'm hard pressed to say we should attempt the work to support that at the moment without through costing.

docs/devspace-proposal/client.md
```diff
@@ -0,0 +1,153 @@
...
+
+The script creates an ingress for `developer.outreach.io` that points to an external service (`web-master.outreach-staging.com`).
+
+- [ ] Remove the localizer orca-proxy pod from client deployment.
+
+### Deploy `client` from Source Code
+
+The deployment mechanism is a bit different between KiND and loft clusters because of how can the images be made available to the cluster. Both variants use Buildkit to build the images, the difference is in where the Buildkit runs. Both variants also use devspace CLI to orchestrate deployment under the hood.
+
+`devenv` passes `$DEPLOY_TO_DEV_VERSION` set to `local` when deploying from source.
```
jaredallard, line 62: We should document all of these env variables. I'm also generally not a huge fan of exposing this because this is become lossy from a strongly typed perspective. That being said, I know we need them. Just providing context on my concerns.

docs/devspace-proposal/client.md
```diff
@@ -0,0 +1,153 @@
...
+
+### `apps dev terminal`
+
+There is no `dev` config in devspace so terminal is not supported.
+
+# Reverse Proxy
+
+Routing traffic dynamically based on current devenv state (whether `client` is deployed into cluster, or running locally). By default, all the traffic goes to cluster (which devenv knows where to find it). 
+
+## `proxy register --host <handled hostname> --target <where the traffic goes>`
```
jaredallard, line 118: Instead of proxy commands we should probably just find a way to make this work with ingress entries and annotations. I don't like making the end user use magic strings to create proxy entries when they will ultimately already need to exist in the devenv in some form

docs/devspace-proposal/client.md
```diff
@@ -0,0 +1,153 @@
...
+* Reads a config file from `~/.local/dev-environment/proxy/mappings.json` with existing registrations.
+* Edit `/etc/hosts` to send mappings to `127.0.0.1`.
+* Starts the reverse proxy (listening on port 443).
+* Listens on chages to mappings (either watching the file, or an internal control plane endpoint that can be called from register/unregister).
+
+> Requires sudo to to edit `/etc/hosts` and listen on port 443
+
+## `devenv` Daemon
+
+`devevn` runs on dev machine as a daemon. This lets us run all the traffic through the revese proxy, switch clusters without having to change `/etc/hosts`, show devs error messages when devenv status isn't quiet right. They don't need to worry whether the tunnel/proxy is running or not. Should include status icon in menu bar.
```
jaredallard, line 143: I've mentioned this before, but doing the daemon isn't really something we should focus on right now. As a north star, sure, it makes sense, but it's a pretty significant work item right now to do.

jaredallard: Not saying it can't be here, just adding that context.

docs/devspace-proposal/client.md
```diff
@@ -0,0 +1,153 @@
...
+
+We can produce notifications when something happens (alert worthy for example). Or in the morning, with nudge to update apps in devenv.
+
+## Dev DNS Server
```
jaredallard, line 147: Important to note this is only really needed if:

a) We don't have public facing DNS entries
b) We modify `/etc/hosts` to avoid roundtrip

docs/devspace-proposal/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+With a proxy running locally, there doesn't need to be a direct connection to cluster as the traffic from https://bento1a.outreach-dev.com/ is directeted by the reverse proxy.
+
+## Flagship Server & Worker
+
+Flagship runs in Kubernetes as well, but being a legacy codebase, built on Ruby on Rails, having more than one "mode" (server, worker, client), being rather large and having long history (meaning .git folder is BIG) presents special needs for runtime. We sync the local code into a container running in kubernetes with gems already built in.
+
+## Giraffe
+
+Giraffe as a Node.js service needs as much performance from a dev machine as we can get. It takes a while to build. It's hard and expensive to match modern ARM Mac performance at building Javascript, and so the best tool available is used - developer's machine. Giraffe is built locally, and then synced to a container in Kubernetes.
```
jaredallard, line 83: Can we define what "building" is here? Are we talking about just TS -> JS transpiling? Also would be good to talk about what the performance of syncing those build artifacts are.

docs/devspace-proposal/README.md
```diff
@@ -0,0 +1,83 @@
+# Development @ Outreach
```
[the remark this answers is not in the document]
plisy: That's a whole different discussion though. Devenv works differently again for k3s. Most of the things are the same between KiND and loft. Other than building docker images, they are pretty much the same.

K3s + Codespaces would have more differences I assume given your starting point is already inside the K8s cluster there.

docs/devspace-proposal/README.md
plisy: Giraffe has its build system. It's mostly TS->JS but more things can be involved. So to build giraffe here, means to run it's build ...

docs/devspace-proposal/client.md
plisy: I don't expect devs to use this command manually. It's primarily for scripting.

docs/devspace-proposal/client.md
plisy: Not all the hosts have public DNS entries (e.g. client doesn't)

docs/devspace-proposal/client.md
plisy: I'm more inclined to pull it out of this repo. Working with concrete services makes more sense here, cause these are more of a how-to-guides for one-of setups, given they rely on "legacy" scripts.

jaredallard: Sorta, but this is also very important for documenting why these features are developed. We need to communicate this to end users outside of Outreach

docs/devspace-proposal/client.md
jaredallard: Sorry, I meant if we don't go the jaredallard-devenv.public.domain.name route

jaredallard: Which we pretty much decided we won't do, was adding context for others

docs/devspace-proposal/README.md
jaredallard: OK but it's good to spell that out because JS is not a compiled language so seeing "copy build artifacts" is a bit confusing without denoting that

docs/devspace-proposal/client.md
plisy: The DNS is kinda aspirational anyway. Doing good enough DNS server is fairly easy, making systems use it is dark magic.

docs/devspace-proposal/README.md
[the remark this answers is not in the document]
jaredallard: Right but part of this entire effort was to figure out our path forward for cloud. If we optimize this for loft then we're committing to loft, at least in the short term, and need to make sure we're OK with that

docs/devspace-proposal/README.md
plisy: There are very few large JS projects without a build step today.

docs/devspace-proposal/bootstrap.md
jaredallard: I'm saying take the stage 1 of our docker images for Go and turn that into its own container that spits out binar(ies) and use that for both the dev base image and in CI. In CI the second stage would be just for deployment, and probably not used in dev. Does that make sense? I can add more if not

docs/devspace-proposal/client.md
plisy: Not saying we need to put it with this proposal, but I do think there's enough DX value there for it to be worth investing in.

docs/devspace-proposal/README.md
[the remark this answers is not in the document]
plisy: I get that, but cloud seems to not be so important. I might be wrong though.

docs/devspace-proposal/bootstrap.md
plisy: Yeah, I didn't really expect this to be the way to go, but maybe someone else reviewing the PR will notice something I didn't.

docs/devspace-proposal/README.md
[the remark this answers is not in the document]
jaredallard: I don't think it was that cloud wasn't important, but that the current cloud was too expensive, so we were to find a route that is cost-effective. We still have a plethora of issues on local machines that the cloud is solving, we can't forget about that pain. If we just stop supporting cloud, those all come back.

cc @deregtd

docs/devspace-proposal/client.md
jaredallard: For sure, I agree with doing it, just that it's not really the highest pri item to do right now.

docs/devspace-proposal/README.md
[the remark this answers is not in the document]
plisy: This proposal still counts on supporting loft.

docs/devspace-proposal/client.md
jaredallard: Maybe one day this will be easier 😭

plisy: I love your optimism 🤣

docs/devspace-proposal/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+We build a few kinds of applications today at Outreach. They tend to have different needs in terms of dev machine setup and actual code runtime. Instead of trying to provide once completely unified dev experience, we try to tailor the tooling for the task, and developer expectations.
+
+This document is a high level overview of how we use `devenv` to build different kinds of applications.
+
+## Bootstrap services
+
+Bootstrap microservices (overwhelming number of repos) run in Kubernetes in production and get their configuration from config maps and secrets. Instead of having local machine friendly way of configuring these, doing service discovery (for dependencies), we support running in Kubernetes during development as well.
```
george-e-shaw-iv, line 9: i don't understand the last sentence of this paragraph

docs/devspace-proposal/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Running in Kubernetes
+
+`devenv` provides two deployment modes
+
+1. pre-built service is useful for testing latest service from git main (or master in some cases) branch
```
george-e-shaw-iv, line 15:
```suggestion
1. released version of the service via the latest pushed docker image into our image repository
```
isn't that more correct?

george-e-shaw-iv: maybe also appending to that - using the deployment manifests from the remote source code repository that correspond to that same release

docs/devspace-proposal/README.md
plisy: Right now, we have special way to configure a service to run locally (the files in `~/.outreach/...`) which doesn't apply to running in Kubernetes. 

We also have special service discovery for running locally.

plisy: With devspace, you won't run service locally anymore.

docs/devspace-proposal/README.md
plisy: sure, let's do that.

docs/devspace-proposal/README.md
george-e-shaw-iv, line 9:
```suggestion
Bootstrap microservices (overwhelming number of repos) run in Kubernetes in production and get their configuration from config maps and secrets. This is opposed to the local machine friendly way of configuring these services which uses a different discovery mechanism for config. For service discovery in bootstrapped applications the same package is used which uses build tags to resolve to different endpoints, depending on the environment it the binary was built in.
```

george-e-shaw-iv: Is that what you were getting at then?

docs/devspace-proposal/README.md
plisy: Do we do that though? I thought we just grab the default branch and deploy using those manifests.

docs/devspace-proposal/README.md
plisy: It took me longer to parse your version, but then I wrote the first one, so I'm biased.

george-e-shaw-iv: yeah im probably adding to much information to mine. i dunno i just didn't understand what was originally being said

docs/devspace-proposal/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+The same environment variables and mounts are available as well in order for the service to have the right configuration for running.
+
+Developers can build and start the service, run tests, including e2e.
+
+#### Performance
+
+The development pods are not going to stay around forever and need to be recreated when manifests and Dockerfile change. With cloud devenv they also need to be recreated after the cluster goes to sleep.
+
+This leads to long time to ready/run in such a container. To avoid this, nightly build creates images with latest packages built in, Buildkit and dev pod share package cache, so devs don't have to redownload them all the time.
```
george-e-shaw-iv, line 54: is this only a problem with the current runtime (loft)? also what does this do, if anything, for performance after the "first run" scenario (once code/deployment/image changes start rolling in)?

plisy: Pre-built containers are a good thing cause they already bundle dependencies. So when you run for the first time, you don't have to go pull them again.

plisy: It's less of a problem in KiND because there's only one node and caching works better.

george-e-shaw-iv: what about other solns to the cloud runtime? like k3s+codespaces?

docs/devspace-proposal/bootstrap.md
```diff
@@ -0,0 +1,93 @@
...
+   - removed resource limits
+   - mounted local volume for caching packages
+   - added `DEV_EMAIL` env variable
+   - added `GH_TOKEN` env variable
+   - added `NPM_TOKEN` env variable
+   - added `VERSION` env variable
+   - set up credentials from env variables to be used for pulling packages (npm, golang, gems)
+   - command tails a log file (this is where application will `tee` logs so they land in Datadog)
+1. `devspace` syncs local source code into the dev container (and keeps syncing them as long as the session is open)
+1. `devspace` forwards ports (8000 for metrics, 8080 for HTTP, 5000 for gRPC, **TBD** for debugger)
```
george-e-shaw-iv, line 64:
```suggestion
1. `devspace` forwards ports (8000 for internal HTTP (metrics, readiness, liveness, etc.), 8080 for public HTTP, 5000 for gRPC, **TBD** for debugger)
```

docs/devspace-proposal/README.md
plisy: You don't necessarily even need to build docker images there. You can set up that thing to behave as if it were a pod in the cluster. Secrets & config maps might need some love, but it's not an unsurmountable issue.

K3s+codespace is a fundamentally different solution to what we have now, cause your dev box is part of the kubernetes cluster in that scenario (like a node in the cluster).

Running things in docker has much less overhead than on mac. I assume some properties of that solution would be awesome, some would be not so great, but unless we go and build a service in such env, we won't know for sure.

plisy: That was probably more general description than needed.

In k3s+codespace this would not be a problem cause the packages would be cached locally - pretty much the same way KiND does it. Except with KiND you have host machine and VM for docker (at least on mac). In codespace, you have just the codespace, so even less indirection.

docs/devspace-proposal/todos.md
```diff
@@ -0,0 +1,34 @@
...
+
+# Bootstrap
+
+- [ ] Add image_registry var to bootstrap jsonnet files. It should only apply to app specific images. Not common ones, like temporal, busybox, etc.
+- [ ] Move `devspace.yaml` to `devbase` for bootstrap services. (Use $DEVENV_APPNAME instead of template from bootstrap for setting the right APP)
+- [ ] `devenv` should set devspace configuration file explicitly. If <repo_root>/devspace.yaml available use it, otherwise fallback to `<repo_root>/.bootstrap/devspace.yaml`.
+- [ ] Set up `.dockerignore` for bootstrap images.
+- [ ] Add loading images into KiND cluster in post-build hook.
+- [ ] When building docker image for loft dev-environment, pre-create Buildkit pod.
+- [ ] Set up remote debugging in dev container.
```
jaredallard, line 17: This is probably post-MVP

docs/devspace-proposal/todos.md
```diff
@@ -0,0 +1,34 @@
...
+- [ ] Send logs from dev container to Datadog. (This most likely needs to be done in devbase devspace.yaml file as the start command for the dev container)
+
+# Flagship
```
jaredallard, line 20: Again we need to not keep adding more flagship/client specific information to this PR, this is all public.

docs/devspace-proposal/todos.md
```diff
@@ -0,0 +1,34 @@
...
+- [ ] Add support for dev of flagship worker
+- [ ] Implement `devenv apps dev terminal` for multiple services.
+- [ ] Add support for dev of flagship kafka-worker. (This might not be needed?)
+- [ ] Add support for dev of flagship console. (This might not be needed?)
+
+# Client
+
+- [ ] Add support for deploy client form source 
+- [ ] Remove the localizer orca-proxy pod from client deployment.
+- [ ] Validate reverse port forwarding through `devspace` for client dev
```
jaredallard, line 33: devspace isn't involved in webapp development.. right? Just the devenv reverse proxy

docs/devspace-proposal/todos.md
plisy: We have the ability to debug code locally. We should probably not lose it.

docs/devspace-proposal/todos.md
```diff
@@ -0,0 +1,34 @@
...
+
+- [ ] Pass `$DEVENV_*` env variables to `devspace`, `scripts/deploy-to-dev.sh`, and `./scripts/devenv-apps-dev.sh`
+- [ ] Implement `devenv apps dev` (and `./scripts/devenv-apps-dev.sh`)
+- [ ] Implement `devenv apps dev stop`
+- [ ] Implement `devenv apps dev terminal`
+- [ ] Implement `devenv apps deploy` using `devspace deploy` (notice how it doesn't take path for deploying from source code. It uses repo root instead)
+
+# Bootstrap
+
+- [ ] Add image_registry var to bootstrap jsonnet files. It should only apply to app specific images. Not common ones, like temporal, busybox, etc.
```
jaredallard, line 11: If this is an extVar it's going to be somewhat complicated and will need to communicate with the COR team -- we don't support adding more of those right now, easily.

docs/devspace-proposal/todos.md
plisy: Well, either that or we go with the reverse port-forwarding option at first like @george-e-shaw-iv suggested. It's easier to implement.

docs/devspace-proposal/todos.md
plisy: Is there really no way to provide a default there?

plisy: There is, but not with extVars. The whole thing needs to be a function.

plisy: So we can either sed the thing and go that way. Or we do something funky with our `jsonnet`.

docs/devspace-proposal/todos.md
jaredallard: We have debugging, via dlv already in devspace. I was thinking about this as supporting it in XYZ editor, which to me is post-MVP still. Correct me if I'm wrong.

plisy: We also have connecting with vscode now.

bhogatemadhura (approved):

PR #213, opened 2022-03-11
feat: Implemented `apps deploy` using `devspace`

plisy:

Implementing `devenv apps deploy` based on the documented logic.

- [ ] Deal with application types as they don't make sense anymore. (knowing things are bootstrapped probably still does, but that can turn into a flag where it makes sense)

## Jira ID

[XX-XX]


plisy:
Will open cleaner PRs easier to review

PR #220, opened 2022-03-16
feat: `registry get` returns gcr.io/outreach-docker for local devenv

plisy:

This is very useful for getting `devspace` working with local devenv (it doesn't get there on it's own, but helps).

## Jira ID

[XX-XX]


plisy:
Oh. I see. Bad idea

plisy:
> Oh. I see. Bad idea

Maybe not

PR #222, opened 2022-03-17
feat: Implemented `apps deploy` and `apps delete` using `devspace`

plisy:

- Adds `apps deploy` using `devspace`
- Adds `apps delete` using `devspace`
- Enables deploying to KiND clusters using `devspace`
- Gated behind `--x-use-devspace` feature flag

## Jira ID

[DTSS-1455]
[DTSS-1451]
[DTSS-1458]

## Notes for your reviewers

[DTSS-1455]: https://outreach-io.atlassian.net/browse/DTSS-1455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DTSS-1451]: https://outreach-io.atlassian.net/browse/DTSS-1451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DTSS-1458]: https://outreach-io.atlassian.net/browse/DTSS-1458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/app/common.go
```diff
@@ -0,0 +1,196 @@
...
+		devspace, err := ensureDevspace(a.log)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to ensure devspace is installed")
+		}
+
+		if err := a.clusterTypeSupported(ctx, devspace, devspaceYamlPath, vars); err != nil {
+			return nil, err
+		}
+
+		// We assume individual profiles don't add dev configs. If they do, this won't work.
```
jaredallard, line 130: We should parse the config as YAML, not regex, this is going to be a lot more likely to break

pkg/app/common.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+		cmd.Env = append(cmd.Env, vars...)
+		return cmd, nil
+	}
+
+	return nil, fmt.Errorf("no fallback script or devspace.yaml found for the application")
+}
+
+func (a *App) clusterTypeSupported(ctx context.Context, devspaceBin, devspaceConfigPath string, envVars []string) error {
+	if a.Local && a.kr.Type == kubernetesruntime.RuntimeTypeLocal && a.Type == TypeBootstrap {
```
jaredallard, line 169: Same thing here, let's parse the config instead. Also, we can shallow return instead of keeping this in an indent.

pkg/app/delete.go
```diff
@@ -46,8 +52,27 @@ func (a *App) deleteBootstrap(ctx context.Context) error {
...
+			"./scripts/deploy-to-dev.sh",
+			"./scripts/devenv-apps-deploy.sh",
+		},
+		fallbackCommandArgs: []string{"delete"},
+	})
+}
+
 func (a *App) Delete(ctx context.Context) error {
 	var err error
+	//nolint:exhaustive // Why: We don't want to delete the app that supports devspace without the x-use-devspace flag.
```
jaredallard, line 75: In this case can we just check it, if not using that flag we return an error?

pkg/app/deploy.go
```diff
@@ -189,30 +193,12 @@ func (a *App) buildDockerImage(ctx context.Context) error {
...
-				return true
-			}
-
-			// filter jobs without our annotation
-			return job.Annotations[DeleteJobAnnotation] != "true"
-		},
-	})
-	if err != nil {
+	var err error
+	if err = a.deleteJobs(ctx); err != nil {
```
jaredallard, line 197:
```suggestion
	if err := a.deleteJobs(ctx); err != nil {
```

pkg/app/deploy.go
```diff
@@ -189,30 +193,12 @@ func (a *App) buildDockerImage(ctx context.Context) error {
...
 		a.log.WithError(err).Error("failed to delete jobs")
 	}
 
+	//nolint:exhaustive // Why: We don't want to delete the app that supports devspace without the x-use-devspace flag.
```
jaredallard, line 201:
```suggestion
        var err error
	//nolint:exhaustive // Why: We don't want to delete the app that supports devspace without the x-use-devspace flag.
```

pkg/app/deploy.go
```diff
@@ -231,3 +217,79 @@ func (a *App) Deploy(ctx context.Context) error { //nolint:funlen
...
+// deployCommand returns the command that should be run to deploy the application
+// There are two ways to deploy:
+// 1. If there's an override script for the deployment, we use that.
+// 2. If there's no override script, we use devspace deploy directly.
+// We also check if devspace is able to deploy the app (has deployments configuration).
+// Skips building images locally if app is already prebuilt (!Local)
+func (a *App) deployCommand(ctx context.Context) (*exec.Cmd, error) {
+	args := []string{"deploy"}
+	if !a.Local {
+		args = append(args, "--skip-build")
```
jaredallard, line 230: What does this flag do? Can we document this?

pkg/app/deploy.go
```diff
@@ -231,3 +217,79 @@ func (a *App) Deploy(ctx context.Context) error { //nolint:funlen
...
+	if err := a.deleteJobs(ctx); err != nil {
+		a.log.WithError(err).Error("failed to delete jobs")
+	}
+
+	cmd, err := a.deployCommand(ctx)
+	if err != nil {
+		return err
+	}
+
+	cmd.Stdin = os.Stdin
```
jaredallard, line 256: nit: We're doing this alot, might be nice to make a function that does this, e.g.:

```go
func useOSStdout(c *exec.Cmd) *exec.Cmd {
  c.Stdin = os.Stdin
  c.Stdout = os.Stdout
  c.Stderr = os.Stderr
}

// usage
usOSStdout(exec.CommandContext(ctx, "command", "arg")).Run()

File: pkg/app/deploy.go [hunk 8]
```diff
@@ -98,7 +102,7 @@ func (a *App) deployBootstrap(ctx context.Context) error { //nolint:funlen
 			},
 			Validator: func(obj *unstructured.Unstructured) bool {
 				var pod *corev1.Pod
-				err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &pod)
+				err := apiruntime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &pod)
```
jaredallard, line 105: Do we need to alias this?

pkg/app/common.go
```diff
@@ -0,0 +1,196 @@
...
+	devspaceVersion := "v5.18.4"
+	devspaceDownloadURL := fmt.Sprintf(
+		"https://github.com/loft-sh/devspace/releases/download/%s/devspace-%s-%s",
+		devspaceVersion,
+		runtime.GOOS,
+		runtime.GOARCH)
+
+	return cmdutil.EnsureBinary(log, "devspace-"+devspaceVersion, "devspace", devspaceDownloadURL, "")
+}
+
```
jaredallard, line 33: These should have comments on the types, all unexported and exported fields (just leaving this one comment for all of them)

pkg/app/app.go
```diff
@@ -24,6 +24,8 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+// TODO: Do we want to redo Type? Given we support more than 1 fallback/override script kinda invalidates it.
```
jaredallard, line 27: `TODO` needs a JIRA ticket, or turn into a `IDEA`, either way format is: `TODO/IDEA(ghusername)`

pkg/app/delete.go
plisy: I want to get rid of the Type altogether. We just need a flag saying it's bootstrap, cause you can make some extra assumptions about bootstrapped apps.

pkg/app/deploy.go
plisy: Not anymore. I needed the `runtime` package here before.

pkg/app/deploy.go
plisy: I really don't think that's any better.

This version is straightforward & dumb. Looking up util func to see whether it's doing something I don't want it to to save 3 lines of code is not.

pkg/app/common.go
```diff
@@ -0,0 +1,196 @@
...
+type devspaceCommandOptions struct {
+	requiredConfig string
+	devspaceArgs   []string
+
+	// If one of these exists, we don't invoke devspace.
+	fallbackCommandPaths []string
+	fallbackCommandArgs  []string
+}
+
+func (a *App) command(ctx context.Context, opts *devspaceCommandOptions) (*exec.Cmd, error) {
```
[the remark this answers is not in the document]
plisy: I'm not exactly sure what you mean by this. But this is one process, to figure out which one it is, we have to check whether there is a config for it, and we have order in which we do it.

I can split it into two funcs if that helps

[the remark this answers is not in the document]
jaredallard: A function is the ligherweight approach to this, basically create a type for each one of these scenarios. Legacy, and devspace. Two functions work as well, but it means we're limited to how much logic for devspace/legacy we can have. Having distinct structs between legacy and devspace allows us to add more functionality per method we support. Think `kubernetesruntime`, if that helps

pkg/app/deploy.go
jaredallard: Except its not 3 lines of code, it's repeated all over the place in this package... I'm not super sure if your argument makes all that much sense given it's essentially the same as what you did for `fileExists`

jaredallard: That being said, this is a nit, but if you feel super strongly about repeating this I'm cool with it.

pkg/app/deploy.go
jaredallard: right, i added the other suggestion. generally you want to declare err in it's own scope to prevent accidental error shadowing. This is a pretty common pattern

pkg/app/deploy.go
plisy: `fileExists` is different in that it abstracts how we check for file (remote/local) so we have single func with list of files to check and different lookup strategies, vs two places with the same list of files to check

pkg/app/deploy.go
jaredallard: Right, but why? Can we add a comment on that branch denoting why we have different behaviour for local? Maybe I missed it.

pkg/app/common.go
jaredallard: You can with `--skip-info`, that prints clean YAML

pkg/app/deploy.go
plisy: I added the comment

pkg/app/common.go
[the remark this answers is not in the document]
plisy: This doesn't fully work, cause you don't have two distinct cases, either all in on overrides, or all in on devspace.

You can have 3 states for each of the commands. 
- override
- devspace
- not supported

And it depends on whether there's override script, or the config exists

pkg/app/common.go
plisy: done

pkg/app/common.go
plisy: Unfortunately, this won't do cause printing config strips profiles, and reading `devspace.yaml` directly may or may not have the profiles at all but have them included from some place else.

plisy: I left a comment about it

pkg/app/common.go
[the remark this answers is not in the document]
plisy: split things into separate funcs

pkg/app/deploy.go
plisy: done

pkg/app/deploy.go
plisy: done

jaredallard (approved):

PR #223, opened 2022-03-17
feat: Added `apps dev` using `devspace`

plisy:

- Support `apps dev` and `apps dev stop` using `devspace`

## Jira ID

[DTSS-1452]

## Notes for your reviewers

[DTSS-1452]: https://outreach-io.atlassian.net/browse/DTSS-1452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/app/dev.go
```diff
@@ -0,0 +1,111 @@
+package app
```
jaredallard, line 1: Same general feedback about attaching to the app struct here, but otherwise this looks pretty good!

Same thing for `TODO` though

plisy: What's the difference with `Deploy` vs `Dev` in terms of attaching to app struct?

plisy: fixed the todos

jaredallard (approved):

PR #231, opened 2022-03-26
feat(devspace): `apps dev --app` sets `$DEVENV_DEV_DEPLOYMENT_PROFILE`

plisy:

Passing `$DEVENV_DEPLOY_APP_PROFILE` to `devspace` lets us start the right binary, replace the right deployment ...

Other PRs to make it work e2e:
- https://github.com/getoutreach/bootstrap/pull/1108
- https://github.com/getoutreach/devbase/pull/161

## Jira ID

[DTSS-1466]

## Notes for your reviewers

[DTSS-1466]: https://outreach-io.atlassian.net/browse/DTSS-1466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/devenv/apps/dev/dev.go
+++ b/cmd/devenv/apps/dev/dev.go
@@ -40,7 +40,7 @@ type Options struct {
 	conf *rest.Config
 
 	// Path is the app to dev
-	AppProfile    string
+	DeploymentProfile    string
 	AppNameOrPath string
 	LocalImage    bool
 	Terminal      bool
@@ -76,8 +76,8 @@ func NewCmd(log logrus.FieldLogger) *cli.Command {
 				Usage: "Open an interactive terminal to the dev container instead of running the application",
 			},
 			&cli.StringFlag{
-				Name:  "app",
-				Usage: "When project has multiple applications, specify which application to dev",
+				Name:  "deployment",
+				Usage: "When project has multiple deployments, specify which deployment to substitute for the dev container",
 			},
 		},
 		Subcommands: []*cli.Command{
@@ -96,10 +96,10 @@ func NewCmd(log logrus.FieldLogger) *cli.Command {
 			o.LocalImage = c.Bool("local-image")
 			o.Terminal = c.Bool("terminal")
 
-			// TODO(DTSS-1466): Add flag validation for profiles
-			appFlag := c.String("app")
-			if appFlag != "" {
-				o.AppProfile = fmt.Sprintf("app__%s", appFlag)
+			// If not set, go with default deployment
+			deploymentFlag := c.String("deployment")
+			if deploymentFlag != "" {
+				o.DeploymentProfile = fmt.Sprintf("deployment__%s", deploymentFlag)
 			}
 			return o.Run(c.Context)
 		},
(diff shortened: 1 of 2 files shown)


george-e-shaw-iv (commented):
My only concern with this is the following: In the current stage of things we're able to set nomenclature to whatever we want. It seems weird that we're refering to a conglomerate as an `app` indirectly via `devenv apps deploy` and now we're referring to specific binaries/targets within the app via `--app`. I don't know what to do here or if it's worth getting pedantic about, but I thought I'd call it out.

Will defer to Jared for final approval on this PR since he's been guiding the UX of the devenv recently.

jaredallard:
Hmmmm, I don't feel as strongly about app as @george-e-shaw-iv does, but I do think he's got a point. Both in that it's confusing nomenclature, to some extent, but also in that it's somewhat against our current service requirements in our "startup in a box". We're pretty strong on "one application per repository", and repositories that have "multiple apps" really have multiple modes, or "profiles". How would we feel about `mode` or `profile` instead?

plisy:
We were planning to do the work to officially support multi-entrypoint. So that's not exactly true.

I think both `profile` and `mode` are more confusing. I'm not claiming `app` is the solution. Though we do use that nomenclature in labels, and the values seem to match binary names.

jaredallard:
Keep in mind the work to support multi-entrypoint is scoped JUST to CDC, it's not going to be global support for multiple entrypoints because we don't allow multiple services in a repository.

plisy:
But we were planning to support multiple tightly related entry points in one repo.

deregtd:
Can we just call it deployment or something?  By default with no deployment specified it does the one bootstrapped one, if you specify an alternate deployment it'll handle that?  mode/profile/app/facet all sound kinda wrong to me.

plisy:
Do we create separate flags for non-deployment scenarios?

jaredallard:
> Can we just call it deployment or something? By default with no deployment specified it does the one bootstrapped one, if you specify an alternate deployment it'll handle that? mode/profile/app/facet all sound kinda wrong to me.

Deployment is fine w/ me, I was trying to think outside of just services but I guess that doesn't make all that much sense because this only matters to services

plisy:
Switched to deployment

cmd/devenv/apps/dev/dev.go
```diff
@@ -38,10 +39,11 @@ type Options struct {
...
 
-	// App is the app to dev
-	App        string
-	LocalImage bool
-	Terminal   bool
+	// Path is the app to dev
+	DeploymentProfile string
+	AppNameOrPath     string
+	LocalImage        bool
+	Terminal          bool
```
jaredallard, line 46: Can you add comments to these fields describing what they are in the standard go comment format?

cmd/devenv/apps/dev/dev.go
```diff
@@ -82,13 +88,19 @@ func NewCmd(log logrus.FieldLogger) *cli.Command {
 			if err != nil {
 				return err
 			}
-			o.App = c.Args().First()
+			o.AppNameOrPath = c.Args().First()
 			// TODO(DTSS-1494) use git to get root directory
-			if o.App == "" {
-				o.App = "."
+			if o.AppNameOrPath == "" {
+				o.AppNameOrPath = "."
```
jaredallard, line 94: Instead of relative paths, would it be better to evaluate the path here and make it absolute?

jaredallard (approved):
LGTM

cmd/devenv/apps/dev/dev.go
plisy: Maybe, but I'm not going to change how this works all across apps commands now.

cmd/devenv/apps/dev/dev.go
plisy: This got auto-merged, will add in the next PR

PR #233, opened 2022-03-28
fix(status): dont add artifical timeout

jaredallard:
## Jira ID

[XX-XX]

### DIFF
--- a/cmd/devenv/status/status.go
+++ b/cmd/devenv/status/status.go
@@ -164,8 +164,7 @@ func (o *Options) GetStatus(ctx context.Context) (*Status, error) {
 		return status, nil
 	}
 
-	timeout := int64(5)
-	_, err := o.k.CoreV1().Pods("default").List(ctx, metav1.ListOptions{Limit: 1, TimeoutSeconds: &timeout})
+	_, err := o.k.CoreV1().Pods("default").List(ctx, metav1.ListOptions{Limit: 1})
 	if err != nil {
 		status.Status = Degraded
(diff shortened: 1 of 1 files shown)


cmd/devenv/status/status.go
```diff
@@ -164,8 +164,7 @@ func (o *Options) GetStatus(ctx context.Context) (*Status, error) {
 		return status, nil
 	}
 
-	timeout := int64(5)
-	_, err := o.k.CoreV1().Pods("default").List(ctx, metav1.ListOptions{Limit: 1, TimeoutSeconds: &timeout})
+	_, err := o.k.CoreV1().Pods("default").List(ctx, metav1.ListOptions{Limit: 1})
```
george-e-shaw-iv, line 167:
```suggestion
	timeout := int64(15)
	_, err := o.k.CoreV1().Pods("default").List(ctx, metav1.ListOptions{Limit: 1, TimeoutSeconds: &timeout})
```

george-e-shaw-iv (approved):
i think it'd be safer to keep a timeout, just triple it or something. approved regardless

cmd/devenv/status/status.go
jaredallard: My problem with this is that k8s somewhat manages its own timeouts so I'd rather not artificially limit it here. I'm not super attached though

george-e-shaw-iv: if they have their own timeouts thats fine

PR #258, opened 2022-04-27
feat(bootstrap): v9.37.0-rc.21

plisy:

CLI telemetry

## Jira ID

[XX-XX]


.releaserc.yaml
```diff
@@ -23,13 +23,3 @@ plugins:
     - assets:
         - "dist/*.tar.gz"
         - "dist/checksums.txt"
-
-  ### IMPORTANT: This needs a block later.
```
jaredallard, line 27: Need to re-add this

PR #268, opened 2022-05-04
fix: rebootstrap cli

george-e-shaw-iv:
## Jira ID

[XX-XX]


.circleci/config.yml
```diff
@@ -1,12 +1,11 @@
 # Please re-run stencil after any changes to this file.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@dev:first
```
george-e-shaw-iv, line 4: worrisome

.circleci/config.yml
```diff
@@ -56,7 +53,9 @@ workflows:
               only:
                 - release
     jobs:
-      - shared/merge_into_release
```
george-e-shaw-iv, line 59: maybe also worrisome i can't tell

george-e-shaw-iv (commented):
lgtm other than the two comments

.circleci/config.yml
plisy: afaik this is fine

.circleci/config.yml
plisy: this is scary, did you gitblame why it's there?

.releaserc.yaml
```diff
@@ -23,13 +23,3 @@ plugins:
     - assets:
         - "dist/*.tar.gz"
         - "dist/checksums.txt"
-
-  ### IMPORTANT: This needs a block later.
-  # Build and push our docker image
```
plisy, line 28: This needs to come back

plisy (commented):
`.releaserc.yaml` needs to be reverted

plisy (approved):
shipit

PR #269, opened 2022-05-04
fix: No forced loft login on `devenv context`

plisy:

Not all devs have access to loft and/or loft license. For these calling context would block things.

## Jira ID

[XX-XX]


cmd/devenv/context/context.go
```diff
@@ -162,8 +162,10 @@ func (o *Options) Run(ctx gocontext.Context) error {
 	clusters := make([]*kubernetesruntime.RuntimeCluster, 0)
 	for _, r := range runtimes {
 		r.Configure(o.log, b)
-		if err := r.PreCreate(ctx); err != nil {
-			o.log.WithError(err).Warnf("Failed to setup runtime %s, skipping", r.GetConfig().Name)
+
+		if isAccessible, err := r.IsAccessible(ctx); err != nil {
+			o.log.WithError(err).Warnf("Failed to check runtime %s availability, skipping", r.GetConfig().Name)
```
george-e-shaw-iv, line 167: this won't skip, it'll continue on in the loop

george-e-shaw-iv (approved):
i think this makes sense

PR #288, opened 2022-05-14
fix: enable pod logging when waiting times out

ChrisPRobinson:
Enables more descriptive logging when waiting for pods to complete fails

## Jira ID

[DTSS-1735]

## Notes for your reviewers

[DTSS-1735]: https://outreach-io.atlassian.net/browse/DTSS-1735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/devenvutil/devenv.go
+++ b/pkg/devenvutil/devenv.go
@@ -171,13 +171,14 @@ func DeleteObjects(ctx context.Context, log logrus.FieldLogger,
 
 // FindUnreadyPods checks all namespaces to find pods that are unready, they are
 // then returned. If an error occurs, err is returned.
-func FindUnreadyPods(ctx context.Context, k kubernetes.Interface) ([]string, error) {
+func FindUnreadyPods(ctx context.Context, k kubernetes.Interface) ([]string, []*corev1.Pod, error) {
 	pods, err := k.CoreV1().Pods(metav1.NamespaceAll).List(ctx, metav1.ListOptions{})
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to list pods")
+		return nil, nil, errors.Wrap(err, "failed to list pods")
 	}
 
-	unreadyPods := []string{}
+	unreadyPodNames := []string{}
+	unreadyPods := []*corev1.Pod{}
 	for i := range pods.Items {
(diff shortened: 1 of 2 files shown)


ChrisPRobinson:
Need to do a final e2e test of this on Monday

pkg/devenvutil/devenv.go
```diff
@@ -209,34 +210,97 @@ func FindUnreadyPods(ctx context.Context, k kubernetes.Interface) ([]string, err
...
+
 	return ctx.Err()
 }
+
+// pod is a wrapper around []*corev1.Pod so that it can marshal to
+// a structured log
+type pods []*corev1.Pod
+
+// MarshalLog implements the log.Marshaler interface for []*corev1.Pod
+func (p *pods) MarshalLog(addField func(key string, v interface{})) {
```
jaredallard, line 260: Does this actually work? We don't use the gobox logger, it'd be cool if it did though

ChrisPRobinson: ugh I made terrible assumptions. the marshaling works but my integration of course does not...

ChrisPRobinson: @jaredallard fixed.

pkg/devenvutil/devenv.go
```diff
@@ -209,34 +210,90 @@ func FindUnreadyPods(ctx context.Context, k kubernetes.Interface) ([]string, err
...
-		log.WithError(err).WithField("pods", unreadyPods).
+		log.WithError(err).WithField("pods", unreadyPodNames).
 			Info("Waiting for pods to be ready")
 
 		async.Sleep(ctx, 30*time.Second)
 	}
 
+	if ctx.Err() != nil {
+		// Write out a bit more detailed info on the prior to exit
+		log.WithError(err).WithField("pods", PodsStateInfo(unreadyPods)).
```
ChrisPRobinson, line 247: It might be better to change the field name and message, given this will be written out after a log.WithError.

jaredallard (approved):

PR #292, opened 2022-05-17
fix(devenv): bump timeout to 20 minutes

jaredallard:

Devenv provisions are getting _pretty_ slow for some of these E2E tests and other environments, so for now I'm bumping the timeout. Longer-term we might need to figure out a better way to handle this.

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/devenvutil/devenv.go
+++ b/pkg/devenvutil/devenv.go
@@ -222,7 +222,7 @@ func FindUnreadyPods(ctx context.Context, k kubernetes.Interface) ([]string, err
 
 // WaitForAllPodsToBeReady waits for all pods to be unready.
 func WaitForAllPodsToBeReady(ctx context.Context, k kubernetes.Interface, log logrus.FieldLogger) error {
-	ctx, cancel := context.WithTimeout(ctx, time.Minute*10)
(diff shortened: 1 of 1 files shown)


ChrisPRobinson (approved):
LGTM

would it make sense to enable this to be parametrizable? Perhaps overridden through an env var if teams need to?

This change also has the impact of for bad builds on circle ci in that places that failed due to reasons will now wait for 20 mins instead of 10 potentially spending more money in CI too. We could even just default it in the make script and if teams need to change it they could. I think this PR is fine just thinking through other options.

jaredallard:
> LGTM
> 
> would it make sense to enable this to be parametrizable? Perhaps overridden through an env var if teams need to?
> 
> This change also has the impact of for bad builds on circle ci in that places that failed due to reasons will now wait for 20 mins instead of 10 potentially spending more money in CI too. We could even just default it in the make script and if teams need to change it they could. I think this PR is fine just thinking through other options.

Yeah fair question, we def need to work on this logic. I'm not super sure what the best options are for that currently. This is purely to solve issues with some developers legitimately having their env take this long to provision.

plisy (approved):

PR #298, opened 2022-05-27
feat: Shim `kubectl` for child processes

plisy:

User machine `kubectl`, devenv Kubernetes, and devenv `kubectl` can be incompatible. This forces use of fixed, embedded `kubectl` by devenv and child processes it spawns.

## Jira ID

[XX-XX]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,11 @@ _ := $(shell ./scripts/devbase.sh)
 
 include .bootstrap/root/Makefile
 
+LDFLAGS += -X k8s.io/component-base/version.gitMajor=1
+LDFLAGS += -X k8s.io/component-base/version.gitMinor=23
+LDFLAGS += -X k8s.io/component-base/version.gitVersion=v1.23.5
+LDFLAGS += -X k8s.io/component-base/version.gitCommit=272114478c66b8250050dd68d4719c46c2ab2088
+
 ###Block(targets)
 .PHONY: e2e-override
 e2e-override:
--- a/Makefile
+++ b/Makefile
@@ -4,12 +4,12 @@ _ := $(shell ./scripts/devbase.sh)
 
 include .bootstrap/root/Makefile
 
+###Block(targets)
 LDFLAGS += -X k8s.io/component-base/version.gitMajor=1
 LDFLAGS += -X k8s.io/component-base/version.gitMinor=23
 LDFLAGS += -X k8s.io/component-base/version.gitVersion=v1.23.5
 LDFLAGS += -X k8s.io/component-base/version.gitCommit=272114478c66b8250050dd68d4719c46c2ab2088
 
-###Block(targets)
 .PHONY: e2e-override
 e2e-override:
 	TEST_OUTPUT_FORMAT=standard-verbose TEST_FLAGS=-v TEST_TAGS=or_e2e ./scripts/shell-wrapper.sh test.sh
--- a/internal/shim/kubectl.embed
+++ b/internal/shim/kubectl.embed
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+# We want to use the devenv that triggered this command. 
+# devenv sets the environment variable $DEVENV_BIN with path to itself.
+DEVENV_BIN="${DEVENV_BIN:-devenv}"
+
+"$DEVENV_BIN" --skip-update kubectl "$@"
--- a/internal/shim/kubectl.go
+++ b/internal/shim/kubectl.go
@@ -19,12 +19,9 @@ var kubectlScript []byte
 //		devenv --skip-update kubectl "$@"
 //
 // The function also adds the script to $PATH so it's picked by child processes.
-func AddKubectl(opts ...func(*Options)) error {
+func AddKubectl(opts ...Option) error {
 	o := &Options{}
-
-	for _, opt := range opts {
-		opt(o)
-	}
+	o.apply(opts)
 
 	devenvShimDir := o.dir
 	if devenvShimDir == "" {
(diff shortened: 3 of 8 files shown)


pkg/app/run.go
```diff
@@ -145,5 +145,5 @@ func (a *App) RunStop(ctx context.Context) error {
 		return errors.Wrap(err, "failed to stop dev mode for the application")
 	}
 
-	return a.appsClient.Set(ctx, &apps.App{Name: a.RepositoryName, Version: a.Version})
+	return nil
```
plisy, line 148: This isn't related, but it needed to happen and was sitting there, waiting for a PR to tag along.

Makefile
```diff
@@ -4,6 +4,11 @@ _ := $(shell ./scripts/devbase.sh)
 
 include .bootstrap/root/Makefile
 
+LDFLAGS += -X k8s.io/component-base/version.gitMajor=1
```
jaredallard, line 7: Can we move this to the `targets` block so it's not overwritten by bootstrap?

internal/shim/kubectl_test.go
```diff
@@ -0,0 +1,23 @@
...
+func TestAddKubectl(t *testing.T) {
+	tempDir := t.TempDir()
+
+	err := AddKubectl(WithShimDir(tempDir))
+	assert.NilError(t, err)
+
+	p, err := exec.LookPath("kubectl")
+	assert.NilError(t, err)
+
+	assert.Assert(t, p == filepath.Join(tempDir, "kubectl"))
```
ChrisPRobinson, line 22: Nit: perhaps run `kubectl version` to ensure the version is what is expected?

internal/shim/kubectl.embed
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# We want to use the devenv that triggered this command. 
+# devenv sets the environment variable $DEVENV_BIN with path to itself.
+DEVENV_BIN="${DEVENV_BIN:-devenv}"
+
+"$DEVENV_BIN" --skip-update kubectl "$@"
```
jaredallard, line 7:
```suggestion
exec "$DEVENV_BIN" --skip-update kubectl "$@"
```

ChrisPRobinson (commented):
LGTM

internal/shim/kubectl.go
```diff
@@ -0,0 +1,64 @@
...
+
+//go:embed kubectl.embed
+var kubectlScript []byte
+
+// AddKubectl creates a kubectl shell script in $HOME/.local/dev-environment/bin. This script executes
+//
+//		devenv --skip-update kubectl "$@"
+//
+// The function also adds the script to $PATH so it's picked by child processes.
+func AddKubectl(opts ...func(*Options)) error {
```
jaredallard, line 22:
```suggestion
func AddKubectl(opts ...func(*Options)) error {
```

nit: Create a `Option` type

internal/shim/kubectl.go
```diff
@@ -0,0 +1,64 @@
...
+	o := &Options{}
+
+	for _, opt := range opts {
+		opt(o)
+	}
```
jaredallard, line 27: I'd recommend making a receiver on `Options` called `apply` that does this.

internal/shim/kubectl_test.go
plisy: That won't work in tests I believe. At least not ones run from VSCode.

ChrisPRobinson (approved):

internal/shim/kubectl_test.go
plisy: Also, that problem would be caught deploying things using devspace.

internal/shim/kubectl.go
```diff
@@ -0,0 +1,64 @@
...
+		f, err := os.OpenFile(kubectlPath, os.O_CREATE|os.O_WRONLY, 0o744)
+		if err != nil {
+			return errors.Wrapf(err, "failed to open file %s", kubectlPath)
+		}
+		defer f.Close()
+
+		_, err = f.Write(kubectlScript)
+		if err != nil {
+			return errors.Wrapf(err, "failed to write to file %s", kubectlPath)
+		}
```
jaredallard, line 54:
```suggestion
		if _, err := f.Write(kubectlScript); err != nil {
			return errors.Wrapf(err, "failed to write to file %s", kubectlPath)
		}
```

nit

internal/shim/kubectl.go
```diff
@@ -0,0 +1,64 @@
...
+		devenvShimDir = filepath.Join(home, ".local", "dev-environment", "bin")
+		err = os.MkdirAll(devenvShimDir, 0o755)
+		if err != nil {
+			return errors.Wrap(err, "failed to create $HOME/.local/dev-environment/bin dir")
+		}
+	}
+
+	kubectlPath := filepath.Join(devenvShimDir, "kubectl")
+	if _, err := os.Stat(kubectlPath); os.IsNotExist(err) {
+		f, err := os.OpenFile(kubectlPath, os.O_CREATE|os.O_WRONLY, 0o744)
```
jaredallard, line 45: Alternativel you could make this shorter by just doing `os.WriteFile()`

internal/shim/kubectl_test.go
ChrisPRobinson: Part of problem you wanted to solve was to ensure the version of kubectl is the right one, no?

jaredallard (approved):
All my comments are nits around style, core is good to me. Happy to chat about them or show suggestions as needed

PR #306, opened 2022-06-10
fix: update deps

jaredallard:

Title says, it does.

## Jira ID

[XX-XX]


george-e-shaw-iv (approved):
k

PR #321, opened 2022-06-24
feat: Added `--with-dependencies` flag to `apps deploy|run`

plisy:

Added the ability to deploy the app and it's dependencies for both `devenv apps deploy` and `devenv apps run`.

Since it's a higher risk change, plan is to roll it out behind a flag at first and after some baking, make it default.

## Jira ID

[DTSS-130]

## Notes for your reviewers

[DTSS-130]: https://outreach-io.atlassian.net/browse/DTSS-130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/app/app.go
```diff
@@ -173,6 +173,35 @@ func NewApp(ctx context.Context, log logrus.FieldLogger, k kubernetes.Interface,
 	return &app, nil
 }
 
+type DevenvConfig struct {
+	Dependencies struct {
```
jaredallard, line 177: nit: struct comment

pkg/app/app.go
```diff
@@ -173,6 +173,35 @@ func NewApp(ctx context.Context, log logrus.FieldLogger, k kubernetes.Interface,
...
+		// Optional is a list of OPTIONAL services e.g. the service can run / gracefully function without it running
+		Optional []string `yaml:"optional"`
+
+		// Required is a list of services that this service cannot function without
+		Required []string `yaml:"required"`
+	} `yaml:"dependencies"`
+}
+
+func (a *App) config() (*DevenvConfig, error) {
+	f, err := os.Open(filepath.Join(a.Path, "devenv.yaml"))
```
jaredallard, line 187: We stopped doing this logic in the e2e runner, we can just use `devenv.yaml` at this point

pkg/app/app.go
```diff
@@ -173,6 +173,35 @@ func NewApp(ctx context.Context, log logrus.FieldLogger, k kubernetes.Interface,
...
+		f, err = os.Open(filepath.Join(a.Path, "service.yaml"))
+	}
+
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read devenv.yaml or service.yaml")
+	}
+	defer f.Close()
+
+	var cfg DevenvConfig
+	if err = yaml.NewDecoder(f).Decode(&cfg); err != nil {
```
jaredallard, line 198:
```suggestion
	if err := yaml.NewDecoder(f).Decode(&cfg); err != nil {
```

pkg/app/deploy.go
```diff
@@ -26,24 +26,76 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+type DeploymentOptions struct {
+	UseDevspace        bool
+	SkipDeployed       bool
+	DeployDependencies bool
```
jaredallard, line 32: nit: comments

jaredallard (approved):

pkg/app/deploy.go
```diff
@@ -26,24 +26,76 @@ import (
...
+	})
+
+	if opts.DeployDependencies {
+		if err := app.deployDependencies(ctx, log, k, b, conf, kr, opts.UseDevspace); err != nil {
+			return err
+		}
+	}
 
-	if useDevspace || forceDevspace {
+	if opts.SkipDeployed {
```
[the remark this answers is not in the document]
plisy: deps skip by default, but the `devenv apps deploy <appName>|<path>` don't. and I'm pretty sure that's the behavior we want here (It's current behavior, the logic behind overwrite what's there vs skip would just get unclear fast, and we'd end up explaining it in support)

PR #333, opened 2022-07-18
feat: cleanup velero namespace

artemchernyak-outreach:

Cleans up the velero namespace after snapshot restore. This should reduce resource use and hopefully allow the loft clusters to scale.

## Jira ID

[DTSS-1724]

## Notes for your reviewers

[DTSS-1724]: https://outreach-io.atlassian.net/browse/DTSS-1724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/devenv/provision/provision.go
+++ b/cmd/devenv/provision/provision.go
@@ -41,15 +41,12 @@ import (
 	"github.com/jetstack/cert-manager/cmd/ctl/pkg/factory"
 	"github.com/jetstack/cert-manager/cmd/ctl/pkg/renew"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	kruntime "k8s.io/apimachinery/pkg/runtime"
 
 	cmclient "github.com/jetstack/cert-manager/pkg/client/clientset/versioned"
 )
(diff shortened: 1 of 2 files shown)


cmd/devenv/provision/provision.go
```diff
@@ -321,6 +321,30 @@ func (o *Options) snapshotRestore(ctx context.Context) error { //nolint:funlen,g
...
+				return false
+			}
+
+			return true
+		},
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed to cleanup namespaces")
+	}
+
```
jaredallard, line 347:
```suggestion
	if err := devenvutil.DeleteObjects(ctx, o.log, o.k, o.r, devenvutil.DeleteObjectsObjects{
		Type: &corev1.Namespace{
			TypeMeta: metav1.TypeMeta{
				Kind:       "Namespace",
				APIVersion: corev1.SchemeGroupVersion.Identifier(),
			},
		},
		Validator: func(obj *unstructured.Unstructured) bool {
			var daemonset *corev1.Namespace
			if err := kruntime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &daemonset); err != nil {
				return true
			}

			if daemonset.Name == "velero" {
				return false
			}

			return true
		},
	}); err != nil {
		return errors.Wrap(err, "failed to cleanup namespaces")
	}

```

nit: try to inline err checks when possible

jaredallard (commented):
Do we want to just uninstall the velero helm chart instead? That should cleanup the resources. That'd just be `kubectl delete helmchart -n kube-system velero` iirc

cmd/devenv/provision/provision.go
```diff
@@ -310,46 +310,20 @@ func (o *Options) snapshotRestore(ctx context.Context) error { //nolint:funlen,g
...
-		},
-		Validator: func(obj *unstructured.Unstructured) bool {
-			var pod *corev1.Pod
-			if err := kruntime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &pod); err != nil {
-				return true
-			}
-
-			for i := range pod.Spec.InitContainers {
-				cont := &pod.Spec.InitContainers[i]
-				if cont.Name == "restic-wait" {
```
artemchernyak-outreach, line 339: Good call @jaredallard on removing the entire helm chart. Looks like it also takes care of the `restic-wait` pods.

jaredallard (approved):

PR #374, opened 2022-09-02
fix: Restore deletion of pods with restic-wait init containers.

jkinkead:
This was deleted in #333, but is still required.


cmd/devenv/provision/provision.go
```diff
@@ -320,6 +323,39 @@ func (o *Options) snapshotRestore(ctx context.Context) error { //nolint:funlen,g
 	}
 	o.log.Info("Deleted velero helmchart")
 
+	// If we don't delete all pods with restic-wait init containers, container
```
jaredallard, line 326: This should be before we delete the velero helm chart -- otherwise it the mutating webhook won't remove the init container.

jkinkead: Hmm. AFAICT the webhook is added the container to the _Pod_, not e.g. the _StatefulSet_.

If deleting after the velero stuff was gone didn't work, the workaround we've used (delete the couchdb pod after it's crash-looping; have it come back up without the init container) wouldn't work at all.

jaredallard: Yeah, fair point

jaredallard (approved):

PR #380, opened 2022-09-08
feat(provision): add DFM kubernetes enabled check

kaldorn:
Adds a check to the docker for mac configuration for whether kubernetes is enabled and if so, disables it.

[DT-207]


cmd/devenv/provision/docker_for_mac.go
```diff
@@ -120,6 +121,13 @@ func reconcileDockerForMacConfig(_ context.Context, settingsFile string) (bool,
 		}
 	}
 
+	if kubernetesEnabled, ok := settings["kubernetesEnabled"].(bool); ok {
+		if kubernetesEnabled {
+			modified = true
+			settings["kubernetesEnabled"] = false
```
jkinkead, line 127: I'm not seeing the schema of this file documented anywhere. Is there a link to some Docker docs you could throw at the top of this function?

Adding a trivial unit test while you're in here wouldn't hurt either (like, verify a file is read / written).

jkinkead (approved):

PR #392, opened 2022-09-27
chore(bootstrap): v11.0.1

jaredallard:

Title says, it does


.tool-versions
```diff
@@ -2,10 +2,10 @@
 golang 1.19
 nodejs 16.13.0
 terraform 0.13.5
-ruby 2.6.6
-protoc 3.19.1
+ruby 2.7.5
+protoc 21.5
```
jaredallard, line 6: @artemchernyak-outreach I _think_ we missed [this commit](https://github.com/getoutreach/stencil-golang/commit/d36162db1df8defc4e7c225e94b695c6cb007de5) to revert, right?

jaredallard:
Something terrible happened here, going to have to investigate it 🤔

.tool-versions
artemchernyak-outreach: Yeah, that definitely looks like a miss.

PR #395, opened 2022-10-17
feat(debug): add debug command

jaredallard:

This PR adds a new `devenv debug` command which returns problems found in the current developer environment. This is done by running "problem" detectors on given Kubernetes resources. Outputted is a user-friendly 'report' with links to different problems (much like a Linter). These pages can be found in the [Wiki](https://github.com/getoutreach/devenv/wiki).

An example of one of these pages: [PodOOMKilled](https://github.com/getoutreach/devenv/wiki/PodOOMKilled)

Example of some fake problems I created (bonus! it found a real problem for me):

devenv debug output

**Note**: This is intentionally limited to being scoped just towards Kubernetes resources currently. Out of scope things are Kubernetes network connectivity and other smoke test like things that'd be handled with sonobouy or some other Kubernetes Cluster acceptance test system (for now).

## Jira ID

[XX-XX]

## Notes for your reviewers

There's totally some optimizations we could make to make this better, but I want to let this go out in it's current form to help enable users to get some insight into their developer environment and focus on optimizations iteratively.


cmd/devenv/debug/problem_pods.go
```diff
@@ -0,0 +1,167 @@
...
+	ID:               "PodImagePullBackOff",
+	ShortDescription: "A pod is in a image pull backoff state, meaning it is unable to pull the image",
+	Detector: func(ctx context.Context, obj runtime.Object) (string, bool, bool) {
+		pod, ok := obj.(*corev1.Pod)
+		if !ok {
+			return "", false, false
+		}
+
+		// Check if the pod has any containers that are in a image pull backoff state
+		for i := range pod.Status.ContainerStatuses {
```
chatchai-outreach, line 81: should we check InitContainerStatuses as well?

cmd/devenv/debug/debug.go
```diff
@@ -0,0 +1,228 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// enabledProblems is a list of problems checkers that are enabled
+var enabledProblems = []Problem{
+	ProblemPodCrashLoopBackOff,
+	ProblemPodNotReady,
+	ProblemPodImagePullBackOff,
+	ProblemPodOOMKilled,
```
chatchai-outreach, line 31: should handle generic PodFailed?

cmd/devenv/debug/problem_pods.go
jaredallard: Oh good call, totally forgot about that!

cmd/devenv/debug/debug.go
jaredallard: Eh -- I feel like I want this to be more specific instead of "catch-alls", PodFailed could be really anything so it's hard to write good docs for what to do for it

chatchai-outreach: what if printing PodFailed with Message, Reason and/or ExitCode and let them continue their investigation or send us these info when they need help from us

cmd/devenv/debug/problem_pods.go
```diff
@@ -0,0 +1,214 @@
...
+	ID:               "PodImagePullBackOff",
+	ShortDescription: "A pod is in a image pull backoff state, meaning it is unable to pull the image",
+	Detector: func(ctx context.Context, obj runtime.Object) (string, bool, bool) {
+		pod, ok := obj.(*corev1.Pod)
+		if !ok {
+			return "", false, false
+		}
+
+		// isImagePullBackOff checks if the container is in a image pull backoff state
+		isImagePullBackOff := func(cs *corev1.ContainerStatus) bool {
```
artemchernyak-outreach, line 97: We should really avoid this pattern.

Ideally, we should put each problem in its own file with helpers clearly broken out. I completely understand that this is an initial implementation, and we can save this for a future refactor.

artemchernyak-outreach (approved):

cmd/devenv/debug/debug.go
jaredallard: The problem with those is they are transitive events, so we can't capture that in `devenv debug` without time windowing. Those will generally be caught by `CrashLoopBackOff` today because that's windowing it for us

PR #398, opened 2022-10-18
fix(app): bootstrap.lock or stencil.lock

jaredallard:

We've started removing `bootstrap.lock` which has broken the app detection logic. Check for `bootstrap.lock` OR `stencil.lock` instead.

## Jira ID

[DT-0]


ChrisPRobinson (approved):
LGTM

jaredallard:
Fixes broke bootstrap release, force merging w/ STAR approver

PR #437, opened 2023-01-11
fix(tf): enable S3 transfer acceleration for snapshots

jaredallard:

This PR enables S3 transfer acceleration for the snapshots bucket in use by Outreach.

## Jira ID

[XX-XX]

## Notes for your reviewers

We should probably move this out one day.


kaldorn (approved):
Accelerate!

PR #438, opened 2023-01-12
feat: S3 Transfer Acceleration support and specifying Kubernetes Version support

jaredallard:
**What this PR does**: This PR adds support for:

 * Specifying Kubernetes version
 * Detect ready pods faster and don't continuously repeat the same message
 * snapshout-uploader: Support aws transfer acceleration endpoints
 * When using an `unstable` release, use the `latest` image in docker to prevent snapshot-uploader failures


cmd/devenv/provision/provision.go
```diff
@@ -635,7 +642,7 @@ func (o *Options) Run(ctx context.Context) error { //nolint:funlen,gocyclo // Wh
 
 	o.log.WithField("runtime", o.KubernetesRuntime.GetConfig().Name).
 		Info("Creating Kubernetes cluster")
-	if err := o.KubernetesRuntime.Create(ctx); err != nil { //nolint:govet // Why: OK w/ err shadow
+	if err := o.KubernetesRuntime.Create(ctx, o.KubernetesVersion); err != nil { //nolint:govet // Why: OK w/ err shadow
```
jkinkead, line 645: Does this really shadow something? Doesn't `:=` _require_ a new lhs target ... ?

cmd/devenv/provision/snapshot.go
```diff
@@ -69,6 +70,12 @@ func (o *Options) stageSnapshot(ctx context.Context, target string, channel box.
 
 	o.log.Info("Waiting for snapshot to finish downloading")
 
+	imageTag := app.Info().Version
+	if strings.Contains(imageTag, "unstable") {
+		// When we have unstable in our version, use the latest tag
```
jkinkead, line 75: Add to PR description?

cmd/devenv/provision/snapshot.go
```diff
@@ -82,9 +89,10 @@ func (o *Options) stageSnapshot(ctx context.Context, target string, channel box.
 					RestartPolicy:      corev1.RestartPolicyOnFailure,
 					Containers: []corev1.Container{
 						{
-							Name:    "snapshot-stage",
-							Image:   "gcr.io/outreach-docker/devenv:" + app.Info().Version,
-							Command: []string{"/usr/local/bin/snapshot-uploader"},
+							Name:            "snapshot-stage",
+							Image:           "gcr.io/outreach-docker/devenv:" + imageTag,
+							ImagePullPolicy: corev1.PullIfNotPresent,
```
jkinkead, line 94: Won't this massively slow down immutable tags?

Can we instead do this only for `latest` or other mutable tags?

cmd/snapshot-uploader/upload.go
```diff
@@ -53,22 +54,42 @@ type SnapshotUploader struct {
...
-func (s *SnapshotUploader) StartFromEnv(ctx context.Context, log logrus.FieldLogger) error {
-	conf := &snapshot.Config{}
-	if err := json.Unmarshal([]byte(os.Getenv("CONFIG")), &conf); err != nil {
-		return errors.Wrap(err, "failed to parse config from CONFIG")
+// isTransferAccelerationURL checks if the url is a transfer acceleration url
+func isTransferAccelerationURL(url string) bool {
+	// Match on:
+	// - .s3-accelerate.amazonaws.com
+	// - .s3-accelerate.dualstack.amazonaws.com
+	return regexp.MustCompile(`\.s3-accelerate(\.dualstack)?\.amazonaws\.com`).MatchString(url)
```
jkinkead, line 62: Make non-capturing:
```suggestion
	return regexp.MustCompile(`\.s3-accelerate(?:\.dualstack)?\.amazonaws\.com`).MatchString(url)
```

internal/e2e/devenv/devenv.go
```diff
@@ -86,6 +86,7 @@ func ProvisionDevenv(t *testing.T, ctx context.Context, opts *ProvisionOpts) fun
 		t.Error(errors.Wrap(err, "failed to load kubernetes runtime"))
 		return nil
 	}
+	popts.KubernetesVersion = "1.21.14"
```
jkinkead, line 89: TODO to take from flag?

pkg/kubernetesruntime/kubernetesruntime.go
```diff
@@ -80,7 +80,7 @@ type Runtime interface {
 	Status(context.Context) RuntimeStatus
 
 	// Create creates a new Kubernetes cluster using this runtime
```
jkinkead, line 82:
```suggestion
	// Create creates a new Kubernetes cluster with the given version using this runtime
```

jkinkead (approved):
Mostly looks great, a few suggestions.

This handles version specification for snapshot generation _only_, correct? There is future work to make it possible to configure the devenv runtime?

cmd/devenv/provision/provision.go
jaredallard: This shadows `err` yes, but after a lot of back and forth we've generally agreed that shadowing `err` is acceptable.

The way I think of it, if you're always doing this where possible, you'll never really have real issues with shadowing `err` because it _should_ be only scoped to this block.

cmd/devenv/provision/snapshot.go
jaredallard: good catch! This was a minor annoyance of mine :D

cmd/devenv/provision/snapshot.go
jaredallard: It won't, this is actually just codifying the behaviour that is already there (when a tag is not `latest` `IfNotPresent` is always used).

You are right here that we actually probably _don't_ want to set this for `latest` though, because that will cause unstable releases to always have an older version

jaredallard:
> Mostly looks great, a few suggestions.
> 
> This handles version specification for snapshot generation _only_, correct? There is future work to make it possible to configure the devenv runtime?

This handles both, you can set it as you generate a snapshot as well as set the version of Kubernetes to use at runtime. Eventually we might want to store that information in the snapshot and dynamically pick based off of that, but this codebase needs quite a bit of help to get there

cmd/devenv/provision/provision.go
jkinkead: I was confused by shadowing vs. reassigning. :)

This is totally acceptable style in my mind.

PR #441, opened 2023-01-13
fix(snapshot): kubernetes-versions -> kubernetes-version

jaredallard:

Smol typo that broke snapshot generation :cry:

## Jira ID

[XX-XX]


kaldorn (approved):
33% less `s`. :shipit:

PR #449, opened 2023-01-17
fix(debug): panic on finding image for container status

jaredallard:

This PR fixes an accidental panic when trying to find an image pull backoff for an init container.

## Jira ID

[DT-3362]

## Notes for your reviewers

[DT-3362]: https://outreach-io.atlassian.net/browse/DT-3362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kaldorn (approved):
That looks reasonable since sl is pod.Spec.Containers.

PR #476, opened 2023-02-21
feat: re-stencil

kaldorn:
## Jira ID

[XX-XX]


monitoring/slos.tf
```diff
@@ -0,0 +1,5 @@
+
+
+// <<Stencil::Block(tfCustomSLODatadog)>>
```
jaredallard, line 3: Yeah, we should skip this for CLIs at least. We should add `{{- $_ := stencil.ApplyTemplate "skipIfNotService" -}}` to the top of the template

jaredallard (approved):
Comment aside, LGTM

PR #477, opened 2023-02-22
feat: Delete crashing pods

marekfexa-outreach:

- After deployment to the local cluster it often happens, that some pods have containers in CrashLoopBackoff state. In that state containers stay for a long time and that slows down mainly devenv provision. This PR is about deleting of such pods, which causes new pods to spawn immediately. 

## Jira ID

[QF-821]

## Notes for your reviewers

[QF-821]: https://outreach-io.atlassian.net/browse/QF-821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jaredallard:
Rebasing to fix tests

pkg/devenvutil/devenv.go
```diff
@@ -361,6 +363,54 @@ func WaitForAllPodsToBeReady(ctx context.Context, k kubernetes.Interface, log lo
 	return nil
 }
 
+func deleteBackoffPods(ctx context.Context, k kubernetes.Interface, log logrus.FieldLogger, pods []*corev1.Pod) {
```
jaredallard, line 366: For all unexported functions, can we have comments describing what they do? We're trying to get better about that in all new code (it's also a requirement of `lintroller` platinum)

pkg/devenvutil/devenv.go
```diff
@@ -361,6 +363,54 @@ func WaitForAllPodsToBeReady(ctx context.Context, k kubernetes.Interface, log lo
...
+	}
+	podNames := mapTo(backoffPods, func(p *corev1.Pod) string { return p.Name })
+	log.Infof("Deleting pods to speed things up: %s", strings.Join(podNames, ", "))
+	if err := deletePodsSync(ctx, k, backoffPods); err != nil {
+		log.Warnf("Unable to delelete crashing pods: %s", err)
+	}
+}
+
+func deletePodsSync(ctx context.Context, k kubernetes.Interface, podsToDelete []*corev1.Pod) error {
+	const gracePeriod int64 = 0
```
jaredallard, line 383: nit: `const` should generally only ever be used for package level variables, also use the default value when able:

```suggestion
	var gracePeriod int64
```

pkg/devenvutil/devenv.go
```diff
@@ -361,6 +363,54 @@ func WaitForAllPodsToBeReady(ctx context.Context, k kubernetes.Interface, log lo
...
+
+	for _, p := range podsToDelete {
+		err := k.CoreV1().Pods(p.Namespace).Delete(ctx, p.Name, *metav1.NewDeleteOptions(gracePeriod))
+		if err != nil {
```
jaredallard, line 387: nit: When checking error, declare it in-line first like so to avoid `err` shadowing:

```suggestion
		if err := k.CoreV1().Pods(p.Namespace).Delete(ctx, p.Name, *metav1.NewDeleteOptions(gracePeriod)); err != nil {
```

pkg/devenvutil/devenv.go
```diff
@@ -361,6 +363,54 @@ func WaitForAllPodsToBeReady(ctx context.Context, k kubernetes.Interface, log lo
...
+			return errors.Wrap(err, "Unable to delete pods")
```
jaredallard, line 388: Go's error standards are to not use capitalization or punctuation in errors

```suggestion
			return errors.Wrap(err, "unable to delete pods")
```

pkg/devenvutil/devenv.go
```diff
@@ -361,6 +363,54 @@ func WaitForAllPodsToBeReady(ctx context.Context, k kubernetes.Interface, log lo
...
+		}
+	}
+	return nil
+}
+
+func filter[T any](slice []T, f func(T) bool) []T {
```
jaredallard, line 394: Given these are the first generics functions being used _at all_ in the DT ecosystem I'd _really_ like these to be well-defined and documented. Most of the developers on the DT team are not familiar with Generics.

(I'm still largely in the boat of _not_ using Generics in Go, the way they were implemented has very little value add in my opinion)

jaredallard: I'd also add that I don't see what Generics are providing here right now -- this is only being used for one type.

I'd suggest moving these into a more general `internal` package to be helpers if that's what you're aiming for, perhaps a new standard library addition into `gobox`?

jaredallard (commented):
Super interesting approach to just delete the pods when checking for unreadiness -- didn't occur to me and seems like a good approach as opposed to having an informer run inside of the cluster doing it for us!

Will hold approval until after a bit of chatting about my other comments, but overall seems good!

pkg/devenvutil/devenv.go
marekfexa-outreach: I was trying last 10 years follow this guidance - https://refactoring.guru/smells/comments (Based on Martin Fowler 1999 refactoring book). But ok, I will describe what code does in textural format.

pkg/devenvutil/devenv.go
marekfexa-outreach: I see, maybe I tried to apply the C# or JS way of doing things here. I will rewrite it to old-school for loop :)

jaredallard:
> Thanks @jaredallard . Comments should be applied. I wonder if we should hide pod deletion behind some environment variable so that it is not on by default (or can be disabled). What do you think?

This seems safe enough to me, so I'm OK with rolling it out as is.

pkg/devenvutil/devenv.go
```diff
@@ -361,6 +363,45 @@ func WaitForAllPodsToBeReady(ctx context.Context, k kubernetes.Interface, log lo
...
+	}
+	log.Infof("Deleting pods to speed things up: %s", strings.Join(podNamesToDelete, ", "))
+	if err := deletePodsSync(ctx, k, podsToDelete); err != nil {
+		log.Warnf("Unable to delelete crashing pods: %s", err)
+		// When the pod deletion does not succeed, we continue
+	}
+}
+
+// gracePeriodForPodDeletion grace period used for deleting pods
+var gracePeriodForPodDeletion int64
```
jaredallard, line 392: nit: I would've kept this in `deletePodsSync`, I don't think it needs to be a package level variable

jaredallard: However, this is unexported, so we can fix it whenever. So, nit!

jaredallard (approved):

PR #480, opened 2023-02-23
feat(devenv): support setting snapshot to use in provision

jaredallard:

This PR adds a `--snapshot` flag that can be used to specifiy the exact snapshot to use, in the combo of `target` and `channel` being provided.

This value can be found with the dtss-tools script:

```
go run ./cmd/dt devenv-snapshots list
```

For internal users :D 

Eventually we should expose functionality to list snapshots in the devenv itself.

## Jira ID

[XX-XX]


george-e-shaw-iv (approved):

deregtd (approved):
Adds ability to select a specific snapshot, which will cause the snapshot discovery to iterate through the available snapshots to find the specified one instead of the default [0 index].

PR #490, opened 2023-03-06
feat: speedup git clone

marekfexa-outreach:
Improves speed of git clone of application dependencies. For bootstrapped application it utilizes git clone of specific git tag, for legacy application it clones just changesets from the last year.

## Jira ID

[QF-836]

## Notes for your reviewers

[QF-836]: https://outreach-io.atlassian.net/browse/QF-836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/app/app.go
```diff
@@ -326,17 +327,29 @@ func (a *App) downloadRepository(ctx context.Context, repo string) (cleanup func
...
+	if a.Type == TypeBootstrap {
+		// Version is always git tag which can be used as a branch
+		//nolint:gosec // Why: On purpose
+		cmd := exec.CommandContext(ctx, "git", "clone", "-b", a.Version, "--single-branch", a.composeGhRepositoryURL(), tempDir)
+		if b, err := cmd.CombinedOutput(); err != nil {
+			return cleanup, errors.Wrapf(err, "failed to shallow clone repository: %s", string(b))
+		}
+	} else {
+		// For legacy application we clone just commits from last year
+		yearAgo := strconv.FormatInt(time.Now().AddDate(-1, 0, 0).Unix(), 10)
```
marekfexa-outreach, line 339: I'm thinking even about monthAgo. Do we expect to deploy to devenv older versions in some use case I don't know about?

pkg/app/app.go
```diff
@@ -326,17 +327,29 @@ func (a *App) downloadRepository(ctx context.Context, repo string) (cleanup func
 
 	a.log.Info("Fetching application")
 
-	//nolint:gosec // Why: On purpose
-	cmd := exec.CommandContext(ctx, "git", "clone", fmt.Sprintf("git@github.com:%s/%s", a.box.Org, a.RepositoryName), tempDir)
-	if b, err := cmd.CombinedOutput(); err != nil {
-		return cleanup, errors.Wrapf(err, "failed to shallow clone repository: %s", string(b))
-	}
+	if a.Type == TypeBootstrap {
```
jaredallard, line 330: Might be safer to check if `Version` is a semantic version instead? Or check if the provided version is a valid reference with Github

pkg/app/app.go
jaredallard: 3 months seems reasonable until specific commits can be checked out in `git`

pkg/app/app.go
```diff
@@ -326,17 +327,29 @@ func (a *App) downloadRepository(ctx context.Context, repo string) (cleanup func
...
-	if b, err := cmd.CombinedOutput(); err != nil {
-		return cleanup, errors.Wrapf(err, "failed to checkout given ref: %s", string(b))
+		//nolint:gosec // Why: On purpose
+		cmd := exec.CommandContext(ctx, "git", "clone", "--shallow-since", yearAgo, a.composeGhRepositoryURL(), tempDir)
+		if b, err := cmd.CombinedOutput(); err != nil {
+			return cleanup, errors.Wrapf(err, "failed to shallow clone repository: %s", string(b))
+		}
+
+		//nolint:gosec // Why: On purpose
+		cmd = exec.CommandContext(ctx, "git", "checkout", a.Version)
```
jaredallard, line 348: You might want to use `git switch` instead, it's supposedly the "better way" to checkout

pkg/app/app.go
marekfexa-outreach: That validation logic is in the function above - https://github.com/getoutreach/devenv/blob/feat/optimize-clone/pkg/app/app.go#L287-L288 . There from there it should be safe to assume that version is semantic (git tag) when type is `bootstrap`

pkg/app/app.go
marekfexa-outreach: Good Idea, updated

pkg/app/app.go
jaredallard: That makes sense, I meant that some applications might not be bootstrap app and still use tags, but if this works :tada:

jaredallard (approved):

pkg/app/app.go
```diff
@@ -446,6 +459,11 @@ func (a *App) determineRepositoryName() error {
 	return nil
 }
 
+// Composes url to GitHub repository for current App
```
jaredallard, line 462: nit: This should follow the `// composeGhRepositoryURL` comment format. eventually lintroller will be in full effect on this repo and be mad about this

pkg/app/app.go
marekfexa-outreach: Then will work the original approach with git clone & checkout without -b flag.  I would go here for simplicity over optimizing these rare cases.

PR #492, opened 2023-03-06
feat(image-resolution): Implement Image resolution 

tonymbugua-outreach:
We want to use "stable" images when deploying applications within devenv (i.e. orange or green channels within maestro

## Jira ID

DT-3506


tonymbugua-outreach:
@jaredallard You mentioned that you would like to have the resolvers implementation in a separate dir but I was running into the import cycle error. Any feedback for a workaround would be greatly appreciated.

pkg/app/githubVersionResolver.go
```diff
@@ -0,0 +1,34 @@
...
+
+func NewGitHubVersionResolver() *GitHubVersionResolver {
+	return &GitHubVersionResolver{}
+}
+
+// GetLatestVersion returns the latest image version based on git repo
+func (g *GitHubVersionResolver) GetLatestVersion(ctx context.Context, log logrus.FieldLogger, app *App) (string, error) {
+	// Find the latest version if not set, or resolve the provided version
+	if app.Version == "" {
+		if err := app.detectVersion(ctx); err != nil {
```
jaredallard, line 23: We should probably keep `app.Version` != "" logic before this, right? Since this should only be invoked if we need to get the latest version.

pkg/app/githubVersionResolver.go
```diff
@@ -0,0 +1,34 @@
...
+			return "", errors.Wrap(err, "failed to determine application version")
+		}
+	} else {
+		// If we had a provided version, attempt to
+		// resolve the version in case it's a branch
+		if err := app.resolveVersion(ctx); err != nil {
```
jaredallard, line 29: I would suggest moving the `resolveVersion` code into this file, since it should be specific to this resolver now.

pkg/app/githubVersionResolver.go
```diff
@@ -0,0 +1,34 @@
...
+type GitHubVersionResolver struct {
+	GitClient *github.Client
+}
+
+func NewGitHubVersionResolver() *GitHubVersionResolver {
+	return &GitHubVersionResolver{}
+}
+
+// GetLatestVersion returns the latest image version based on git repo
+func (g *GitHubVersionResolver) GetLatestVersion(ctx context.Context, log logrus.FieldLogger, app *App) (string, error) {
```
jaredallard, line 20: It looks like `*App` is the reason for the import cycle, I suggest just passing the appName in instead to deviate from the original design. This will prevent an import cycle and be much cleaner.

pkg/app/maestroVersionResolver.go
```diff
@@ -0,0 +1,96 @@
+package app
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/getoutreach/gobox/pkg/box"
+	"github.com/getoutreach/maestro/pkg/mclient"
```
jaredallard, line 9: This is a private repo, so we can't clone in that library unfortunately. We'll need to either:

a) Pull it out and use it
b) Duplicate the logic

b seems like the better option for now since we can just make an HTTP request. We might want to make a ticket to do `a` in the long run.

pkg/app/maestroVersionResolver.go
```diff
@@ -0,0 +1,96 @@
...
+type MaestroVersionResolver struct{}
+
+// GetLatestVersion returns the latest image version based on maestro API
+func (m *MaestroVersionResolver) GetLatestVersion(ctx context.Context, log logrus.FieldLogger, app *App) (string, error) {
+	// get config information from devenv.yaml
+	// get authentication information for maestro
+	// loop through the channels specified in config
+	// if the channel returns 404 retry
+
+	log = log.WithFields(logrus.Fields{
```
jaredallard, line 25: Would it be a good idea to change the logger being passed into these to set `app.name` and `app.versionResolver` automatically?

pkg/app/maestroVersionResolver.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	maestroClient := mclient.NewMaestroClient(maestroClientConfig)
+	foundVersion := ""
+
+	// loop through all the channels
+	for _, channel := range versionResolver.Config.Maestro.Channels {
+		maestroVersion, err := maestroClient.FetchActionable(app.RepositoryName, channel)
+		if err != nil {
+			log.Info(fmt.Sprintf("unable to fetch actionable version for channel %s Err: %s", channel, err.Error()))
```
jaredallard, line 83:
```suggestion
			log.WithField("channel", channel).WithError(err).Info("unable to fetch actionable version")
```

jaredallard: I might be slightly wrong on the syntax here, so might not work as-is

pkg/app/resolver.go
```diff
@@ -0,0 +1,45 @@
...
+	// get resolvers from box config
+	versionResolvers := app.box.DeveloperEnvironmentConfig.VersionResolvers
+
+	// loop through all the enabled versionResolver providers
+	for _, versionResolverType := range versionResolvers.Enabled {
+		// convert to VersionResolverType type
+		versionResolverType := VersionResolverType(versionResolverType)
+
+		// get config implementation based on type
+		config, err := app.getVersionResolverConfigFn(versionResolverType)
```
jaredallard, line 21: It might be a better idea to use a `New<Resolver>(b *box.Config)` function instead that passes in the box config. This would allow the package to determine what fields to grab off of the box config. This could be its config as well as potentially other fields it may need (e.g., vault for maestro)

pkg/app/resolver_test.go
```diff
@@ -0,0 +1,261 @@
+package app
```
jaredallard, line 1: love to see the test coverage! ❤️

pkg/app/versionResolver.go
```diff
@@ -0,0 +1,39 @@
+package app
```
jaredallard, line 1: General comment for all the files, we're missing the copyright and some exported types don't have comments. It'd be great to get that added!

pkg/app/mockVersionResolver.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+type MockVersionResolver struct {
```
jaredallard, line 11: Slight nit, I like to build mock providers that store data instead of explicit `ReturnsErr` options. This makes it a little more natural to use (since If you operate on what you think it'd return, it's easier to determine if it should return an error or not).

Here's an example: https://github.com/getoutreach/argocd/blob/master/maestro-dv-syncer/internal/argocd/argocdtest/argocdtest.go

pkg/app/versionResolver.go
```diff
@@ -0,0 +1,39 @@
...
+	VersionResolverMockString VersionResolverType = "mockVersionResolver"
+)
+
+type VersionResolver interface {
+	GetLatestVersion(context.Context, logrus.FieldLogger, *App) (string, error)
+}
+
+// StringToVersionResolverConfigMap converts a VersionResolverType to a
+// VersionResolver config if found. Returns an error if not found
+var StringToVersionResolverConfigMap = map[string]VersionResolver{
```
jaredallard, line 25: I would suggest using `VersionResolverType` as the key type here and converting the user provided string. Also, `Type` at the end of it feels not idiomatic since it includes the type in the name, I suggest just going with `VersionResolverName`

```
config, found := StringToVersionResolverConfigMap[VersionResolverName(key)]
```

pkg/app/githubVersionResolver.go
tonymbugua-outreach: Moved the `resolveVersion` function to the githubVersionResolver.go file

pkg/app/maestroVersionResolver.go
tonymbugua-outreach: Duplicated mclient into devenv and opened a [JIRA ticket](https://outreach-io.atlassian.net/browse/DT-3571) to track it

pkg/app/maestroVersionResolver.go
tonymbugua-outreach: Updated

pkg/app/maestroVersionResolver.go
tonymbugua-outreach: logger being passed in has the app.name and app.versionResolver set automatically now

pkg/app/versionResolver.go
tonymbugua-outreach: added to all files

pkg/app/versionResolver.go
tonymbugua-outreach: updated map to use `VersionResolverName` as the key and changed datatpe from VersionResolverType to VersionResolverName

pkg/app/githubVersionResolver.go
tonymbugua-outreach: The function interface relies on the app struct to exist to reference some variables within it inorder for maestro to work. Not sure if this needs a rewrite then or we should leave it as is?

pkg/app/resolver.go
tonymbugua-outreach: Sounds good. I will work on trying to refactor this

pkg/app/githubVersionResolver.go
tonymbugua-outreach: Hmm, i didnt change anything about the `app.Version` check from the [main branch](https://github.com/getoutreach/devenv/blob/main/internal/app/app.go#L154). Don't we want to get the latest version when there is no version specified?

tonymbugua-outreach: not sure if i understand the ask here 😅

pkg/app/mockVersionResolver.go
tonymbugua-outreach: Hmm, I spent a few trying to figure out what to store as data but the only data that gets returned is the version (string) and error if its unable to get the version from that version resolver. Maybe we can chat about this more in the meeting I added for Monday next week

tonymbugua-outreach:
```
type VersionResolver interface {
	GetLatestVersion(context.Context, logrus.FieldLogger, *App) (string, error)
}
```

pkg/app/githubVersionResolver.go
tonymbugua-outreach: already done

pkg/app/githubVersionResolver.go
tonymbugua-outreach: Fixed in call today

pkg/app/resolver.go
tonymbugua-outreach: fixed in call today

internal/app/resolvers/github/github.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2023 Outreach Corporation. All Rights Reserved.
+
+// Description: This file has the package github.
+
+// Package github handles the github implementation for VersionResolver.
+package github
```
jaredallard, line 6:
```suggestion
// Copyright 2023 Outreach Corporation. All Rights Reserved.

// Description: See package comment.

// Package github implements the VersionResolver interface using Github as the backing
// version store. Both commits and tags are supported as a source of truth for the latest
// version of an application.
//
// For tags, semantic version constructs are used. The latest semantic version is used.
// For commits, the latest commit in the default branch (determined through the Github API)
// is used.
//
// See`detectVersion` for more information about the behaviour and how to select
// commits and tags for cases where both may be in use.
package github
```

internal/app/resolvers/github/github.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/sirupsen/logrus"
+	"golang.org/x/oauth2"
+)
+
+type VersionResolver struct {
+	boxConfig *box.Config
+}
+
+func New(b *box.Config) types.VersionResolver {
+	return &VersionResolver{
```
jaredallard, line 25:
```suggestion
// VersionResolver implements the VersionResolver interface.
type VersionResolver struct {
	boxConfig *box.Config
}

// New returns a new Github VersionResolver implementation
func New(b *box.Config) types.VersionResolver {
	return &VersionResolver{
```

internal/app/resolvers/github/github.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/getoutreach/gobox/pkg/box"
+	githubauth "github.com/getoutreach/gobox/pkg/cli/github"
+	"github.com/google/go-github/v47/github"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/oauth2"
+)
+
+type VersionResolver struct {
+	boxConfig *box.Config
```
jaredallard, line 21: nit: It's more idiomatic to use `b` instead of `boxConfig` here

internal/app/resolvers/maestro/maestro.go
```diff
@@ -0,0 +1,96 @@
+// Copyright 2023 Outreach Corporation. All Rights Reserved.
+
+// Description: This file has the package maestro.
+
+// Package maestro contains the implementation of maestro for VersionResolver
```
jaredallard, line 5:
```suggestion
// Copyright 2023 Outreach Corporation. All Rights Reserved.

// Description: See the package comment.

// Package maestro contains a VersionResolver implementation using Maestro, and internal
// service in use at Outreach, as the source of truth for the latest version of applications.
```

internal/app/resolvers/maestro/maestro.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// VersionResolver version resolver that implements version retriever from maestro
+type VersionResolver struct {
+	boxConfig *box.Config
+}
+
+func New(b *box.Config) types.VersionResolver {
```
jaredallard, line 27:
```suggestion
// New returns a new Maestro VersionResolver
func New(b *box.Config) types.VersionResolver {
```

internal/app/resolvers/mock/mock.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2023 Outreach Corporation. All Rights Reserved.
+
+// Description: This file implements mock client.
+
+// Package mock contains the implementation of mock for VersionResolver
```
jaredallard, line 5:
```suggestion
// Copyright 2023 Outreach Corporation. All Rights Reserved.

// Description: See package comment.

// Package mock implements a VersionResolver implementation for use in testing.
// This implementation returns static data only and is not meant to be used outside
// of testing environments.
```

internal/app/resolvers/types/types.go
```diff
@@ -0,0 +1,23 @@
...
+
+// VersionResolverName just wraps string which contains the version resolver name
+type VersionResolverName string
+
+// VersionResolver is the interface that is used by resolver to retrieve image version
+type VersionResolver interface {
+	GetLatestVersion(context.Context, logrus.FieldLogger, string, string) (string, error)
+}
+
+type VersionResolverCreator func(*box.Config) VersionResolver
```
jaredallard, line 23:
```suggestion

// VersionResolverCreator is the standard format of a `New` function
// that return a new instance of an implementation of the `VersionResolver`
// interface.
type VersionResolverCreator func(*box.Config) VersionResolver
```

internal/app/versionResolver.go
jaredallard, line 1: This file feels like belongs in the `resolvers` folder instead in potentially something like `resolvers.go`. It'll also compose better there.

internal/app/versionResolver.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// MaestroVersionResolverString is the VersionResolverType type for github
+	GitHubVersionResolverString types.VersionResolverName = "github"
+
+	// VersionResolverMockString used to mock version resolver when testing
+	VersionResolverMockString types.VersionResolverName = "mockVersionResolver"
+)
+
+// StringToVersionResolverConfigMap converts a VersionResolverType to a
+// VersionResolver config if found. Returns an error if not found
```
jaredallard, line 29:
```suggestion
// VersionResolver config if found.
```

Maps don't return errors if a key doesn't exist, so slight adjustment there.

pkg/embed/manifests/pre-restore/cert-manager.yaml
jaredallard, line 1: Accidental commit?

internal/app/resolver.go
```diff
@@ -0,0 +1,57 @@
...
+	// loop through all the enabled versionResolver providers
+	for _, versionResolverType := range versionResolvers.Enabled {
+		log.WithFields(logrus.Fields{
+			"app.versionResolver": versionResolverType,
+		})
+		// convert to VersionResolverType type
+		versionResolverType := types.VersionResolverName(versionResolverType)
+
+		// get config implementation based on type
+		config, err := app.getVersionResolverConfigFn(versionResolverType)
```
jaredallard, line 33: This isn't really the config, at this point, since it's the resolver itself. I suggest renaming this to `resolver` and updating the names. This should just be `getVersionResolverByName`.

internal/app/resolver.go
```diff
@@ -0,0 +1,57 @@
...
+func (app *App) HandleVersionResolver(ctx context.Context, log logrus.FieldLogger) error {
+	log = log.WithFields(logrus.Fields{
+		"app.name": app.RepositoryName,
+	})
+
+	// get resolvers from box config
+	versionResolvers := app.box.DeveloperEnvironmentConfig.VersionResolvers
+
+	// loop through all the enabled versionResolver providers
+	for _, versionResolverType := range versionResolvers.Enabled {
```
jaredallard, line 25: A lot of the variable names in this function are long. I'd suggest dropping redundant terms where it makes sense. For example, this function is named `HandleVersionResolver`. We can probably drop `versionResolver` from being a prefix on most of these.

jaredallard (commented):
Left some comments.

Overall, we're close! I'd like to see the maestro client ripped out, though; it's currently leaking many implementation details about a private service. The original design called for just using a standard Golang HTTP client since we only make a single call. Let's continue doing that. It would also reduce the LOC introduced in this PR a lot.

internal/app/resolvers/github/github.go
jaredallard: (`box` is also acceptable)

internal/app/versionResolver.go
tonymbugua-outreach: Tried to move it but the import cycle error comes back again. App is importing the resolvers package and the resolver_test.go file is importing the app package 😢

tonymbugua-outreach:
> Left some comments.
> 
> Overall, we're close! I'd like to see the maestro client ripped out, though; it's currently leaking many implementation details about a private service. The original design called for just using a standard Golang HTTP client since we only make a single call. Let's continue doing that. It would also reduce the LOC introduced in this PR a lot.

Sounds good! Refactored and removed the client. Changed to use API

pkg/embed/manifests/pre-restore/cert-manager.yaml
tonymbugua-outreach: reverted

internal/app/resolvers/maestro/client/client.go
jaredallard, line 1: You should be able to replace this file with `maestroClient.go`, right?

internal/app/resolvers/maestro/client/maestroClient.go
```diff
@@ -0,0 +1,75 @@
...
+	body, err := io.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	var version Version
+	err = json.Unmarshal(body, &version)
+	if err != nil {
+		return nil, err
+	}
```
jaredallard, line 72: nit: You can avoid reading the response into memory here with `json.NewDecoder(resp.Body).Decode(&version)`

internal/app/resolvers/maestro/client/version.go
```diff
@@ -0,0 +1,17 @@
...
+// Description: See the package comment.
+
+// Package client contains implementation for maestro client
+package client
+
+import "time"
+
+// Version is the struct payload returned from maestro actionableVersion
+type Version struct {
+	// we are making the explicit decision that 0 is not a valid id and want to omit it from json
```
jaredallard, line 12:
```suggestion
```

I don't feel like this comment adds much value

internal/app/resolvers/maestro/client/client.go
tonymbugua-outreach: This was the [error](https://app.circleci.com/pipelines/github/getoutreach/devenv/2096/workflows/f8125f5b-5939-4869-94f2-99c305da446e/jobs/5962) coming out of circleCI

jaredallard: Ah, yes, I meant rename `maestroClient.go` to `client.go` because it's already in a descriptive path

tonymbugua-outreach: oh i see. Yes

PR #494, opened 2023-03-07
feat: Dont clone app when deployment will be skipped

marekfexa-outreach:
- Speed up devenv apps deploy due to not cloning out dependencies when there will be no 
- E.g. in searchstore app, there was performed clone of flagship 3x (zero times after this change)

## Jira ID

[QF-838]

## Notes for your reviewers

[QF-838]: https://outreach-io.atlassian.net/browse/QF-838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/app/app.go
```diff
@@ -163,17 +164,6 @@ func NewApp(ctx context.Context, log logrus.FieldLogger, k kubernetes.Interface,
...
 	app.log = app.log.WithField("app.version", app.Version)
 
-	// Download the repository if it doesn't already exist on disk.
-	if app.Path == "" {
-		if err := app.downloadRepository(ctx, app.RepositoryName); err != nil {
-			return nil, errors.Wrap(err, "failed to download repository")
-		}
-	}
-
-	if err := app.determineRepositoryName(); err != nil {
```
marekfexa-outreach, line 173: This call is not needed. Why read app name from `service.yaml` when you use the name of in `devenv` CLI or in dependent application `devenv.yaml`. We always assume that the GH project name is the same as the name in `service.yaml`. Correct?

pkg/app/deploy.go
```diff
@@ -71,6 +71,13 @@ func Deploy(ctx context.Context, log logrus.FieldLogger, k kubernetes.Interface,
 		}
 	}
 
+	// Download the repository if it doesn't already exist on disk.
```
marekfexa-outreach, line 74: This is the gist -> application will be cloned only when it is obvious that source code is needed for actual deployment to devenv

pkg/app/app.go
```diff
@@ -477,3 +494,16 @@ func (a *App) Close() error {
 
 	return nil
 }
+
+// createGhClient creates new github client
+func createGhClient(ctx context.Context) (*github.Client, error) {
```
tomassusovsky-outreach, line 499: can be inlined also in `resolveVersion` and `determineTypeRemote`

pkg/app/deploy.go
tomassusovsky-outreach: as downloadRepository is creating tempDit with timestamp suffix, won't there be an issue, if I call devenv deploy apps --with-deps multiple time, it would be cloned multiple times if needed to be cloned the first time?

pkg/app/app.go
jaredallard: There might be some nuance here with local development since `app.Name` is `.` in that case. Not 100% sure if that was the intention of this

pkg/app/app.go
```diff
@@ -189,16 +179,48 @@ type DevenvConfig struct {
...
-		return nil, errors.Wrap(err, "failed to read devenv.yaml or service.yaml")
+		return nil, errors.Wrapf(err, "failed to get file from GitHub of %s", ConfigFileName)
+	}
+	content, err := config.GetContent()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get content from GitHub of %s", ConfigFileName)
+	}
+
+	var cfg DevenvConfig
+	if err := yaml.NewDecoder(strings.NewReader(content)).Decode(&cfg); err != nil {
```
jaredallard, line 208: Can `config` be passed to `yaml.NewDecoder` directly here, instead of reading it into memory?

pkg/app/app.go
jaredallard: Does https://pkg.go.dev/github.com/getoutreach/gobox/pkg/cli/github#NewClient work instead?

pkg/app/deploy.go
jaredallard: I'm really confused about how this works, how are we able to skip cloning the application? Is this done by skipping it if we've already deployed the application? I'm confused

pkg/app/app.go
marekfexa-outreach: Yeah, that would be nice, but github client `RepositoryContent` type does not implement `io.Reader` interface.

pkg/app/app.go
marekfexa-outreach: yeah, I wanted to use the extract function refactoring of these 8 lines of code. I was hesitant to copy-paste the whole block, I prefer to give it a name and use. But yeah, Jared, use function that was already written is even better :)

pkg/app/deploy.go
marekfexa-outreach: Correct, when the application is already deployed - e.g flagship from the big snapshot there's no need to clone it.  The deployment will then be skipped anyway. In this case, it just gets config (devenv.yaml) using GH API to check that dependencies are already there.

marekfexa-outreach: In other words I'm moving git clone to right before app deployment, where it is obvious that deployment will happen. Current state was clone always.

pkg/app/app.go
marekfexa-outreach: Fixed

pkg/app/deploy.go
marekfexa-outreach: When I was analyzing this change again, I discovered that I completely ignored other usages of `newApp` that will now probably fail due to not having cloned repo. I will rething this change little bit.

pkg/app/app.go
tomassusovsky-outreach: would  replacing`RepositoriesService` `GetContents` call with `DownloadContents` method downloading specified file and returning `io.ReadCloser` interface to the (config) file solve this recommendation?

marekfexa-outreach: Cool, thanks @tomassusovsky-outreach , this will simplify things!

jaredallard (approved):

PR #512, opened 2023-03-17
refactor: Abstract deployment logic

marekfexa-outreach:
- Extract logic related to deployment to Custom (formally legacy) or Bootstrap to separate interface to improve code structure

## Jira ID

[XX-XX]


internal/app/config.go
```diff
@@ -0,0 +1,78 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+
+// Description: This file contains reading and parsing of devenv config
+
+package app
```
marekfexa-outreach, line 5: This is code extraction from app.go file, no change inside

internal/app/app.go
```diff
@@ -63,6 +62,9 @@ type App struct {
 	// Type is the type of application this is
 	Type Type
 
+	// Abstraction over Custom or Bootstrap deployment
```
jaredallard, line 65: nit: Should start with the name of the field

```suggestion
	// Deployment is an abstraction over Custom or Bootstrap deployment
```

internal/app/deployment.go
```diff
@@ -0,0 +1,15 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+
+// Description: This file app functions related to app deployment
+
+package app
+
+import "context"
+
```
jaredallard, line 8: nit: Missing comment on interface

internal/app/bootstrapDeployment.go
jaredallard, line 1: I would suggest making structure like so:

```
deployments/<type>/<type.go>
```

Then use `<type>` (.e.g., `custom`) as the name of the package.

jaredallard: You can do it this way as well, I just like the separation of package space (easier to test and easier to read IMO)

jaredallard:
@marekfexa-outreach Can this be closed?

jaredallard:
Re-open if needed!

PR #515, opened 2023-03-20
refactor: move pkg to internal

marekfexa-outreach:
- move all but snapshoter and embed out of pkg directory
- no other changes made

## Jira ID

[XX-XX]


jaredallard (approved):
You could also get away with moving `embed`, I'd think

PR #518, opened 2023-03-21
refactor: Extract devenv config

marekfexa-outreach:
- Extract separate devenv config related logic to separate package

## Jira ID

[XX-XX]


internal/app/config/local.go
```diff
@@ -0,0 +1,49 @@
...
+		return "", errors.Wrapf(err, "failed to read %s", ServiceConfigFileName)
+	}
+
+	// conf is a partial of the configuration file for services configured with
+	// bootstrap (stencil).
+	var conf struct {
+		Name string `yaml:"name"`
+	}
+
+	if err = yaml.Unmarshal(b, &conf); err != nil {
```
jaredallard, line 44: Since I see it right now

```suggestion
	if err := yaml.Unmarshal(b, &conf); err != nil {
```

internal/app/config/local.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+)
+
+// ReadLocalDevenv returns DevenvConfig from local filesystem
+func ReadLocalDevenv(path string) (*DevenvConfig, error) {
```
jaredallard, line 15: Maybe it makes more sense to expose a generic fetcher for local vs remote and have remote/local implement these through interfaces? Not super attached, but just wonder if it makes sense for us to do this now or not.

marekfexa-outreach: Yeah, I had it like that but refactor it to this, since it felt simpler and doing the same thing

jaredallard (approved):

PR #519, opened 2023-03-21
refactor: repository logic to new package

marekfexa-outreach:
- Extract git repository related logic to separate package to reduce complexity inside app pacakge

## Jira ID

[XX-XX]


jaredallard (approved):
This is way better!

PR #521, opened 2023-03-23
feat: Add snapshot validation commands

Erik-Outreach:

Add snapshot validation commands.

Overall plan:
- Get this merged. Nothing is changed by default in `snapshot generate` and `provision` commands. 
- Next, make changes in `devenv-snapshots` to implement validation there in CircleCI commands:
```
UPLOAD_TO_LOCAL_SNAPSHOT_STORAGE=true devenv snapshot generate --channel rc
devenv destroy
PROVISION_FROM_LOCAL_SNAPSHOT_STORAGE=true devenv provision --snapshot-channel rc
// check that the pods are up and running
devenv snapshot upload --channel rc
```
- Probably delete UPLOAD_TO_LOCAL_SNAPSHOT_STORAGE variable, making `snapshot generate` always generate to local snapshot storage and not upload to final directory.

## Jira ID

[QF-844]

## Notes for your reviewers

[QF-844]: https://outreach-io.atlassian.net/browse/QF-844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/devenv/snapshot/snapshot.go
```diff
@@ -587,3 +640,175 @@ func (o *Options) CreateSnapshot(ctx context.Context) (string, error) {
 		}
 	}
 }
+
```
jaredallard, line 643: nit: function comment

cmd/devenv/snapshot/snapshot.go
```diff
@@ -587,3 +640,175 @@ func (o *Options) CreateSnapshot(ctx context.Context) (string, error) {
...
+		Bucket: &o.b.DeveloperEnvironmentConfig.SnapshotConfig.Bucket,
+		Key:    lockFilePath,
+		Body:   bytes.NewReader(byt),
+	})
+	if err != nil {
+		return err
+	}
+
+	// finally, delete snapshot infrastructure
+	err = snapshotstaging.CleanUp(ctx, o.log)
```
jaredallard, line 742: nit: return output of function instead of assigning `err` that's immediately returned

pkg/snapshotstaging/snapshotstaging.go
```diff
@@ -0,0 +1,187 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
```
jaredallard, line 1: Can any of this be reused by the e2e test? Since it's by-in-large copied from there it'd be nice if we had it working for both cases

cmd/devenv/snapshot/snapshot.go
```diff
@@ -587,3 +640,175 @@ func (o *Options) CreateSnapshot(ctx context.Context) (string, error) {
...
+
+	if _, ok := destLockFile.TargetsV2[name].Snapshots[channel]; !ok {
+		destLockFile.TargetsV2[name].Snapshots[channel] = make([]*box.SnapshotLockListItem, 0)
+	}
+}
+
+func uploadFromMinioToS3(ctx context.Context, source *minio.Client, sourceBucketName, key string, o *Options, s3c *s3.Client) error {
+	resp, err := source.GetObject(ctx, sourceBucketName, key, minio.GetObjectOptions{})
+
+	if err != nil {
```
jaredallard, line 795: nit: no space between err assign and check

```suggestion
	resp, err := source.GetObject(ctx, sourceBucketName, key, minio.GetObjectOptions{})
	if err != nil {
```

jaredallard: Same on L808

cmd/devenv/snapshot/snapshot.go
```diff
@@ -587,3 +640,175 @@ func (o *Options) CreateSnapshot(ctx context.Context) (string, error) {
...
+	// finally, delete snapshot infrastructure
+	err = snapshotstaging.CleanUp(ctx, o.log)
+	return err
+}
+
+func (o *Options) LoadLockFiles(ctx context.Context,
+	source *minio.Client,
+	sourceBucketName string,
+	lockFilePath *string, s3c *s3.Client) (sourceLockFile, destLockFile *box.SnapshotLock, err error) {
+	sourceLockFile = &box.SnapshotLock{}
```
jaredallard, line 750: In general using named return values isn't done that often in Go, it's really easy to accidentally not sent them and they are easy to shadow/use inappropriately.

This function in general would benefit from stronger typing by using structs containing params and not having named return values.

I'd suggest putting `source`, `sourceBucketName` and `lockfilePath` into a struct named `LoadLockfileOptions` that contained the relevant fields, it'd make this a lot more readable. I don't like the usage of named return vars here because it's not that many. The length of the names is also not idiomatic, I'd expect it to be `src` and `dest` because the types are self-explanatory.

pkg/snapshotstaging/snapshotstaging.go
Erik-Outreach: I'm having a bit of trouble re-using the log functions, I'm missing a wrapper which I could use in both cases

cmd/devenv/snapshot/snapshot.go
Erik-Outreach: That makes sense. The named value was removed with a bit of refactor, by deleting minio client and using generic s3 client instead.

pkg/snapshotstaging/snapshotstaging.go
```diff
@@ -0,0 +1,288 @@
...
+func execOSOut(cmd *exec.Cmd) *exec.Cmd {
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd
+}
+
+type SnapshotOptions struct {
+	log logrus.FieldLogger
+	b   *box.Config
+	ctx context.Context
```
Erik-Outreach, line 48: is it controversial to store context in a struct? I'm asking because I don't see it being stored like this anywhere else in this code repo

cmd/devenv/snapshot/snapshot.go
```diff
@@ -165,86 +170,88 @@ func NewCmdSnapshot(log logrus.FieldLogger) *cli.Command {
...
+				Flags: []cli.Flag{
+					&cli.StringFlag{
+						Name:  "channel",
+						Value: string(box.SnapshotLockChannelRC),
+						Usage: "Which channel this snapshot should be uploaded to",
+					},
+					&cli.StringFlag{
+						Name:  "kubernetes-version",
+						Usage: "Kubernetes version to use when generating snapshots. Has no effect if a devenv already exists",
+						Value: "v1.21.14",
```
jaredallard, line 186: This is a mismatch now, I think we're on 1.22.15 now, check `provision.go`

cmd/devenv/snapshot/snapshot.go
```diff
@@ -165,86 +170,88 @@ func NewCmdSnapshot(log logrus.FieldLogger) *cli.Command {
...
+					o.KubernetesVersion = strings.TrimPrefix(c.String("kubernetes-version"), "v")
+
+					return o.Upload(c.Context, s, box.SnapshotLockChannel(c.String("channel")))
+				},
+			},
+		},
+	}
 }
 
+// Spins up devenv, deploys applications and generates a snapshots for all targets, and finally and uploads them.
```
jaredallard, line 220: Go function comments should all begin with the function name, e.g. `// Generate creates a new snapshot ...`

jaredallard: This applies to all the functions in this PR (minus one!)

cmd/devenv/snapshot/snapshot.go
```diff
@@ -587,3 +584,158 @@ func (o *Options) CreateSnapshot(ctx context.Context) (string, error) {
...
+	} else {
+		o.log.WithError(err).
+			Warn("Failed to fetch existing remote snapshot lockfile, will generate a new one")
+	}
+
+	return lockFile, nil
+}
+
+// To make sure the lock file contains the path to the target and channel, avoiding nil pointer exceptions
+func assureLockFilePath(lockFile *box.SnapshotLock, target string, channel box.SnapshotLockChannel) {
```
jaredallard, line 700: This function feels wrongly named. It'd be more common to find a function like `NewLockfile` that would return `box.SnapshotLock` that's fully initialized. I'd suggest doing that here or in `gobox` (since the code to interact with the lock file technically lives there).

I'd suggest changing `LoadLockfile` to return a new lock file if it's not found. Returning `nil, nil` will almost always cause panics down the line.

Erik-Outreach: I was thinking to name this function `assertSnapshotLockListItem` but I had the idea to return the object. It probably best now as long as updating the returned pointer is fine.

cmd/devenv/snapshot/snapshot.go
```diff
@@ -587,3 +574,157 @@ func (o *Options) CreateSnapshot(ctx context.Context) (string, error) {
...
+	} else {
+		o.log.WithError(err).
+			Warn("Failed to fetch existing remote snapshot lockfile, will generate a new one")
+	}
+
+	return lockFile, nil
+}
+
+// getSnapshotLockListItem safely returns SnapshotLockListItem from a SnapshotLock by target and channel, avoiding nil pointer exceptions
+func getSnapshotLockListItem(lockFile *box.SnapshotLock, target string, channel box.SnapshotLockChannel) *[]*box.SnapshotLockListItem {
```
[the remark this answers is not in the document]
Erik-Outreach: that was intentional, I sent the reference back to be updated (so you can update the snapshot item which lives in the snapshot lock object). I was a creative alternative to the previous change set where we needed separate function to assert existence of the "path" to the lock item.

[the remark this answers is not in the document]
Erik-Outreach: I've simplified the logic and removed the extra *

jaredallard (approved):

PR #530, opened 2023-03-29
fix: skip portforwarding by default

zaquestion:


jaredallard (approved):
❤️ perfect, thank you!

PR #532, opened 2023-03-29
fix: increase cert-manager readiness probe delay

kaldorn:
60s -> 80s.

This also adds a job that checks for cert-manager to fully start up so that we can utilize a `kubectl wait -n cert-manager --for=condition=complete job/cert-manager-startupapicheck` command when spinning up dev-environments with devbase.

## Jira ID

[XX-XX]


pkg/embed/manifests/pre-restore/cert-manager.yaml
```diff
@@ -11,12 +11,15 @@ spec:
   set:
     installCRDs: "true"
   valuesContent: |-
+    ## Enable this job to run that completes and is removed immediately after cert-manager is completely ready to go.
+    startupapicheck:
```
jaredallard, line 15: If helm will wait for this, maybe we don't need to bump the initial delay at all?

kaldorn: Does helm block on it?

jaredallard (approved):

PR #539, opened 2023-04-03
feat: delete pods max 3 times for the same owner

marekfexa-outreach:
- Speedup optimization of pod deletion in CrashLoopBackoff can make diagnosing startup problems hard
- This change is about deleting just three times for the same owner reference
- Based on this feedback https://outreach-hq.slack.com/archives/CN9MU7GLW/p1680196696180199

## Jira ID

[QF-853]

## Notes for your reviewers

[QF-853]: https://outreach-io.atlassian.net/browse/QF-853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/devenvutil/devenv.go
```diff
@@ -374,10 +375,38 @@ func hasJobOwner(pod *corev1.Pod) bool {
...
+	}
+
+	val, ok := deletionCounter[ownerName]
+	if !ok {
+		deletionCounter[ownerName] = 1
+		return 1
+	}
+
+	incremented := val + 1
+	deletionCounter[ownerName] = incremented
```
jaredallard, line 400: I think this can be simplified to `deletionCounter[ownerName]++`

internal/devenvutil/devenv.go
```diff
@@ -374,10 +375,38 @@ func hasJobOwner(pod *corev1.Pod) bool {
 	return false
 }
 
-// deleteBackoffPods Filters pods using ProblemPodCrashLoopBackOff detector and call deletePodsSync on them
-func deleteBackoffPods(ctx context.Context, k kubernetes.Interface, log logrus.FieldLogger, pods []*corev1.Pod) {
+// getOwnerName returns owner name for the first owner reference
+func getOwnerName(pod *corev1.Pod) string {
+	for _, owner := range pod.OwnerReferences {
+		return owner.Name
```
jaredallard, line 381: Should we also include the type here, if it's available? Otherwise I think things could collide

internal/devenvutil/devenv.go
marekfexa-outreach: Updated

internal/devenvutil/devenv.go
marekfexa-outreach: I'm not sure if that is needed, since owner reference names seem unique (e.g. templating-service-849485444c). Prefixing it with Kind (e.g ReplicaSet) does not seem better.

jaredallard: A better scenario is a statefulsets, where you could have this collide.

I think it's important to ensure we have kind + namespace + name for anything that's globally scoped.

jaredallard: UID works, I didn't realize that was in the owner reference field!

jaredallard (approved):

PR #540, opened 2023-04-03
fix(cert-manager): wait for webhook to be ready

jaredallard:

This PR changes the devenv to wait for cert-manager to be ready when first provisioning. This should unbreak applications that depend on cert-manager to be in an healthy state when the devenv is provisioned.

## Jira ID

[XX-XX]


cmd/devenv/provision/provision.go
```diff
@@ -534,11 +535,47 @@ func (o *Options) runProvisionScripts(ctx context.Context) error {
...
+	}
+
+	copts.Factory = &factory.Factory{
+		RESTConfig: k8sConf,
+		CMClient:   CMClient,
+		KubeClient: o.k,
+	}
+
+	o.log.Info("Waiting for cert-manager to be ready ...")
+	copts.Run(ctx)
```
jaredallard, line 563: This doesn't return an error, so I'm actually super unsure how to tell if this failed

kaldorn (approved):
This seems like a reasonable thing to try. We can always pivot if necessary in the future.

PR #545, opened 2023-04-04
fix(vpa): remove VPA CRDs for 1.22 compat

jaredallard:

This PR removes the CRDs for the Vertical Pod Autoscaler. This is no longer in use at Outreach, but also the v1 versions [don't work provided by VPA themselves](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/deploy/vpa-v1-crd-gen.yaml), and I don't want to convert them for them, so we are dropping it. Note we've never supported the VPA, just not _erroring_ when VPA resources are created.

## Jira ID

[XX-XX]


kaldorn (approved):
:shipit:

chatchai-outreach (approved):
should we also remove VPA from [jsonnet-lib](https://github.com/getoutreach/jsonnet-libs/blob/456efd5ea2f77ce34bc9261701a99dcdc5e75736/kubernetes/kube.libsonnet#L537-L547) as well?

jaredallard:
> should we also remove VPA from [jsonnet-lib](https://github.com/getoutreach/jsonnet-libs/blob/456efd5ea2f77ce34bc9261701a99dcdc5e75736/kubernetes/kube.libsonnet#L537-L547) as well?

That's a good question, that's really just a fork of the bitnami jsonnet-libs... I don't know if we should continue to modify it or not. I'd imagine one day we'll align with their updated version.

PR #561, opened 2023-04-18
feat: Sync dependency calculation with devbase

marekfexa-outreach:
## Jira ID

[XX-XX]


internal/app/dependencies.go
```diff
@@ -0,0 +1,96 @@
...
+	for k := range deps {
+		keys = append(keys, k)
+	}
+	return keys, nil
+}
+
+// collectTransitiveDependencies grabs required dependencies recursivelly into deps map
+func collectTransitiveDependencies(ctx context.Context, log logrus.FieldLogger, b *box.Config, depName string, deps map[string]struct{}) {
+	// outreach specific hack
+	if depName == "flagship" {
```
jaredallard, line 78: We might be able to remove this hack now, we stopped allowing top level `flagship` awhile ago. If you don't want to remove it now, I'd suggest adding a `TODO` to remove it here (and maybe a log warning that the transitive dependency is using the wrong dependency name).

jaredallard (approved):
Approving with the assumption that my comment will be addressed (on PTO soon, don't want to accidentally block)

internal/app/dependencies.go
marekfexa-outreach: Added as part of the second PR https://github.com/getoutreach/devenv/pull/568/commits/30842612d39826658068847b8c406a7e8a5bed6e

PR #571, opened 2023-04-25
feat: adds flag to leave k8s cluster intact when restore fails & configure timeout

evadnoob:
Adds flag to ignore snapshot restore failures


cmd/devenv/provision/provision.go
```diff
@@ -708,24 +716,29 @@ func (o *Options) Run(ctx context.Context) error { //nolint:funlen,gocyclo // Wh
...
+				if err2 != nil {
+					o.log.WithError(err).Error("failed to remove intermediate environment")
+					return err2
+				}
 
-			return errors.Wrap(err, "failed to provision from snapshot")
+				return errors.Wrap(err, "failed to provision from snapshot")
+			} else {
+				o.log.Warn("")
+				o.log.WithError(err).Warn("failed to provision from snapshot, destroy devenv is disabled")
```
jaredallard, line 740:
```suggestion
				o.log.WithError(err).Warn("failed to provision from snapshot, leaving devenv intact (--destroy-on-failure is false)")
```

cmd/devenv/provision/provision.go
```diff
@@ -708,24 +716,29 @@ func (o *Options) Run(ctx context.Context) error { //nolint:funlen,gocyclo // Wh
...
+			}
```
jaredallard, line 741: I think we should just return here. I don't think we want to do anything else if we failed.

jaredallard: (this is for `destroy-on-failure` being false)

evadnoob: ok, sounds good, was debating that too

jaredallard (approved):

cmd/devenv/provision/provision.go
```diff
@@ -111,7 +121,7 @@ func (o *Options) MarshalLog(addField func(key string, v interface{})) {
 	addField("devenv.provision.snapshot_channel", o.SnapshotChannel)
 	addField("devenv.provision.base", o.Base)
 	addField("devenv.provision.use_devspace", o.UseDevspace)
-
+	addField("devenv.provision.destroy_on_failures", o.DestroyOnFailure)
```
evadnoob, line 124:
```suggestion
	addField("devenv.provision.destroy_on_failure", o.DestroyOnFailure)
```

PR #579, opened 2023-05-01
feat: set kube-reserved cpu and ensure guaranteed qos for control plane pods

kaldorn:
This PR sets aside 500m for the kube daemons running, and patches all control plane pods to ensure they have a QoS of guaranteed. Ideally this PR helps with the resource constraints from large/resource intensive dev environments.

## Jira ID

[XX-XX]


pkg/embed/config/kind.yaml
```diff
@@ -40,3 +40,8 @@ nodes:
         ---
         kind: KubeletConfiguration
         maxPods: 150
+        ---
+        kind: InitConfiguration
```
jkinkead, line 44: Add a link somewhere to where this config is defined?

jkinkead (approved):

PR #581, opened 2023-05-03
feat: ensure control plane pods have guaranteed qos and reserve cpu for kube daemons 

kaldorn:
This PR patches the control plane pods (etcd, kube-controller-manager, kube-scheduler, kube-apiserver) to have a QoS class of guaranteed. This means that they will be the last pods to be preempted when resources become scarce. 

Setting kube-served resources means that k8s won't even see that CPU and kubelet + daemons on the node will have a dedicated amount of resources available.

This _should_ provide some stability to the dev environment when services that are CPU intensive are deployed.

## Jira ID

[DT-3713]

## Notes for your reviewers
To test pull this branch down locally and run 
`make build && ./bin/devenv provision`

[DT-3713]: https://outreach-io.atlassian.net/browse/DT-3713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/kubernetesruntime/kind.go
```diff
@@ -154,6 +162,32 @@ func (kr *KindRuntime) Create(ctx context.Context, kubernetesVersion string) err
 	return errors.Wrap(cmd.Run(), "failed to run kind")
 }
 
+// Writes patch files to devenv directory
```
jkinkead, line 165: Can you link to the docs for this (the patch functionality)? It would also be great to pull a little more docs from the body here, specifically noting that this copies the files from `pkg/embed/config`.

Also, the style guide says it should be `// writePatches writes patch files ...`.

internal/kubernetesruntime/kind.go
```diff
@@ -154,6 +162,32 @@ func (kr *KindRuntime) Create(ctx context.Context, kubernetesVersion string) err
...
+func (*KindRuntime) writePatches(homeDir string) (string, error) {
```
jkinkead, line 166: Please write unit tests for this.

internal/kubernetesruntime/kind.go
```diff
@@ -154,6 +162,32 @@ func (kr *KindRuntime) Create(ctx context.Context, kubernetesVersion string) err
...
+	// Create the subdirectory for patches in the ~/.config/devenv/patches.
+	// We mount that to the KIND control plane.
+	patchesDir := filepath.Join(homeDir, ".outreach", ".config", "dev-environment", "patches")
+	if err := os.MkdirAll(patchesDir, 0o755); err != nil {
+		return "", err
+	}
+
+	// Write all the patch files
+	for _, patchFile := range patchFiles {
+		file := fmt.Sprintf("config/%s", patchFile)
```
jkinkead, line 179: Can you move these to `pkg/embed/patches` instead? That would be less confusing.

pkg/embed/config/etcd.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
```
jkinkead, line 1: Can you throw a one-line comment at the top of all of these resource files noting the point of the patch (reserving resources)?

jkinkead (commented):
Mostly LGTM, apart from missing unit tests. :)

internal/kubernetesruntime/kind.go
```diff
@@ -154,6 +162,33 @@ func (kr *KindRuntime) Create(ctx context.Context, kubernetesVersion string) err
 	return errors.Wrap(cmd.Run(), "failed to run kind")
 }
 
+// writePatches takes the files in /pkg/embed/patches and writes them to a local directory to be mounted to the control-plane
```
jkinkead, line 165: Avoid run-on:
```suggestion
// writePatches takes the files in /pkg/embed/patches and writes them to a local directory to be mounted to the control-plane.
```

jkinkead (approved):
Nice!

PR #588, opened 2023-05-10
fix: move two v1beta1 ingress usages to v1

reseul:
Local-app redirection of orca didn't work due to this obsolete usage.
Updated to v1, and similarly the outlook case (haven't tested that, hope it works)

## Jira ID

[XX-XX]


kaldorn (approved):
That looks right. Approved.

PR #593, opened 2023-05-16
feat: use outreach coredns

kaldorn:
This tells kind/kubeadm to use gcr.io/outreach-docker/coredns:1.10.1 instead of registry.k8s.io/coredns:1.8.7 (or similar, can't fully recall).

It also allows us the ability to use external plugins with coredns.

## Jira ID

[DT-3741]

## Notes for your reviewers

[DT-3741]: https://outreach-io.atlassian.net/browse/DT-3741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/embed/config/kind.yaml
```diff
@@ -1,5 +1,11 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
+kubeadmConfigPatches:
+  - |
+    kind: ClusterConfiguration
+    dns:
+      imageRepository: gcr.io/outreach-docker
```
jkinkead, line 7: This is going to make this even less usable outside of Outreach, but that's fine.

jkinkead (approved):

PR #600, opened 2023-05-19
split generate and upload commands

Erik-Outreach:

It will allow us to change snapshot generation in the follow way:
1 - Deploy all applications per defined targets (SKIP_GENERATE_SNAPSHOTS=true)
2 - Wait extra time for replication to smartstore to complete
3 - Start taking snapshots only after all replication is done (SKIP_DEPLOY_APPS=true)

Blocks this PR: https://github.com/getoutreach/devenv-snapshots/pull/99


jkinkead:
The new `deployTarget` function is doing a lot more than just waiting for apps to boot; it's also running deploys and snapshot commands. Is that the right new default behavior? It doesn't really match the PR description ("Start taking snapshots only after all replication is done").

jaredallard:
Note that we'll need this to get to stable before devenv-snapshots#99 can be merged (it uses rc for PRs, but on merge it uses stable)

jaredallard (approved):

jaredallard:
@Erik-Outreach does this still make sense?

Erik-Outreach:
> @Erik-Outreach does this PR still make sense to keep open, or can it be closed?

Let's close for now, might not be needed

PR #610, opened 2023-05-26
fix(devenv): reenable delibird

jaredallard:
Updates `gobox` to get a stdin fix and reenables delibird.


kaldorn (approved):
turn on delibird

PR #611, opened 2023-05-26
fix: update control-plane resource allocations

jaredallard:
Updates a few control-plane components' resources. The most important
one here is 'etcd', which is really close to these on avg, which isn't
capturing the `max` usage.


tonymbugua-outreach (approved):
LGTM

kaldorn (approved):
Bump em

PR #612, opened 2023-05-26
fix(dfm): 4 -> 5 CPU, memory 12GB -> 16GB, remove old config value

jaredallard:
Changes the number of CPUs passed to the devenv from 4 -> 5 (I've been
running with 5 for a long time with no issues, I was pretty sure this
was our default not sure how I got 5 :D). Bumps RAM from 12GB -> 16GB
because we're at ~85% usage of a base devenv deploy and have plenty of
RAM on user's machines today.

Removes `legacyCgroupV1` config value because we've been on a version 
that supports the newer cgroups (v2) for awhile now.


cmd/devenv/provision/docker_for_mac.go
```diff
@@ -56,8 +56,9 @@ func reconcileDockerForMacConfig(_ context.Context, settingsFile string) (bool,
 		return false, err
 	}
 
-	recommendedCPU := 4
-	recommendedMemory := 12288
+	recommendedCPU := 5
```
jaredallard, line 59: I don't know if we _need_ to bump th CPU by default, but given most users seem to need to do so for more deploys, I felt 5 CPU was a healthy amount. I've also been using this to test some of our heavy deployments to see how it impacts stability.

kaldorn (commented):
Is this what shows as allocatable on the node in dev-env?

jaredallard:
> Is this what shows as allocatable on the node in dev-env?

yessir

jkinkead:
One note from @aocole this morning: He runs at 6 CPU / 16G memory for his work, and has things stable for weeks before there is a problem.

jaredallard:
> One note from @aocole this morning: He runs at 6 CPU / 16G memory for his work, and has things stable for weeks before there is a problem.

So far I'm having success with just 5 CPU on Ben's scenario... if I run into issues I'll bump it to 6, but I'm worried about the impact on Intel mac.s

kaldorn (approved):

kaldorn:
Mine has been at 5 and I'm not sure what happened to do that. Keep in mind with the 500m I set for kube-reserved (kubelet + whatever daemons run on the devenv pod) this is 4.5 cpu allocatable.

PR #621, opened 2023-05-31
feat: Make snapshot generation more efficient by deploying apps simultaneously

Erik-Outreach:


cmd/devenv/apps/deploy/deploy.go
```diff
@@ -94,11 +93,6 @@ func NewCmd(log logrus.FieldLogger) *cli.Command {
 			nArg := c.NArg()
 			if nArg == 0 {
 				return fmt.Errorf("missing appName")
-			} else if nArg > 1 {
-				return fmt.Errorf(
-					"expected at most one argument; got %q",
```
Erik-Outreach, line 99: I think the original implementation was missing this

jaredallard (approved):

cmd/devenv/apps/deploy/deploy.go
jaredallard: It was likely because the original work was done before that was added and merged cleanly.

I removed it in my PR that duplicated the work Chris did (which got merged instead of mine, when I realized). Interesting edgecase, honestly!

PR #622, opened 2023-05-31
fix: update local-path-provisioner to work on arm64

jaredallard:
Updates local-path-provisioner to no longer use our fork and to work on
arm64. This should be the last container in use that does not work on
arm64 natively (important for us using arm64 Linux laptops). Documents
why we need local volume support, now that upstream has it, and updates
the manifests as required.

Also changes a KinD image to match all the other KinD images.


jkinkead:
Are there any performance or stability improvements this gets us, or is this just cleanup?

jaredallard:
> Are there any performance or stability improvements this gets us, or is this just cleanup?

Cleanup + linux/arm64 support (matters for my Gentoo box)

kaldorn (approved):

PR #624, opened 2023-06-01
feat: ensure that vault-secrets-operator has guaranteed qos

kaldorn:
This should ensure that vault-secrets-operator has a QoS of Guaranteed.
It also sets an appropriate priority class.

Looking at peaks from the last 3mo of metrics, there's a handful that have a max more than 128mb. I'm willing to have this OOMKill because it should restart quickly and it _only_ needs to be running when a service is deployed with a new vaultSecret object.

## Jira ID

Part of [DT-3770]

## Notes for your reviewers

[DT-3770]: https://outreach-io.atlassian.net/browse/DT-3770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/embed/manifests/pre-restore/vault-secrets-operator.jsonnet
```diff
@@ -25,6 +25,17 @@ local manifests = {
       vault: {
         address: vault_addr,
       },
+      priorityClassName: 'system-cluster-critical',
+      resources: {
+        limits: {
+          cpu: '100m',
+          memory: '128Mi',
+        },
```
jaredallard, line 33:
```suggestion
           limits: self.requests,
```

nit: alias to requests?

jaredallard (approved):

PR #632, opened 2023-06-05
feat: patch coredns with guaranteed qos and priorityclass

kaldorn:
Since the devenv uses `kubecfg update` to apply the pre-restore manifests, this actually works to patch coreDNS to have guaranteed QoS and priorityclass.

## Jira ID

[DT-3770]

## Notes for your reviewers

[DT-3770]: https://outreach-io.atlassian.net/browse/DT-3770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/embed/manifests/pre-restore/coredns.yaml
```diff
@@ -0,0 +1,27 @@
...
+  template:
+    metadata:
+      labels:
+        k8s-app: kube-dns
+    spec:
+      containers:
+        - name: coredns
+          resources:
+            limits:
+              cpu: 100m
```
jaredallard, line 22: Seems slightly low, maybe 200m?

kaldorn: To set this I looked at metrics of cpu usage for coredns over the last 3mo. In the devenv >95% of usage is under 50m. There are 4 spikes over 100m.

jaredallard (approved):

PR #635, opened 2023-06-07
feat(update): add alert to update to notify users args are not allowed

tonymbugua-outreach:
`Devenv update` should not take in service name as a param to update. Apps that need to be updated should be done so by running `devenv apps deploy <APP>@<VERSION>`

## Jira ID

DT-3791


cmd/devenv/apps/update/update.go
```diff
@@ -120,22 +117,10 @@ func (o *Options) Run(ctx context.Context) error {
...
-				break
-			}
-		}
-		if foundApp == nil {
-			return fmt.Errorf("Unknown app '%s', please use devenv apps deploy first", o.AppName)
-		}
-
-		// only update the app provided
-		deployedApps = []store.App{*foundApp}
+		o.log.Warn("DEPRECATED: Use `devenv apps deploy` instead to update a services version")
```
jaredallard, line 122: I imagine we want to keep the original behaviour and just add the warn+sleep, right?

jaredallard:
@tonymbugua-outreach Mind adding a description of this change and why we're doing it? (should be in the ticket, but if not let me know and we can work on it together!)

cmd/devenv/apps/update/update.go
```diff
@@ -120,7 +118,11 @@ func (o *Options) Run(ctx context.Context) error {
 		return err
 	}
 
+	// devenv apps update should not take any arguments and should use devenv apps deploy instead
 	if o.AppName != "" {
+		o.log.Warn("DEPRECATED: Use `devenv apps deploy` instead to update a services version")
+		time.Sleep(1 * time.Second)
```
jaredallard, line 124:
```suggestion
		time.Sleep(5 * time.Second)
```

I like to make intentional sleeps a little longer :D

jaredallard (approved):

PR #636, opened 2023-06-08
feat(provision): enable setting timeout from environmental variable

kaldorn:
This should make it easier to configure this in CircleCI jobs.

## Jira ID

[XX-XX]


cmd/devenv/provision/provision.go
```diff
@@ -232,7 +233,16 @@ func NewCmdProvision(log logrus.FieldLogger) *cli.Command {
 			o.SnapshotChannel = box.SnapshotLockChannel(c.String("snapshot-channel"))
 			o.Snapshot = c.String("snapshot")
 			o.DestroyOnFailure = c.Bool("destroy-on-failure")
-			o.ProvisionTimeout = c.Duration("provision-timeout")
+
+			if os.Getenv("PROVISION_TIMEOUT") != "" {
```
jaredallard, line 237: You should be able to set `EnvVar` on the flag[1].

[1]: https://pkg.go.dev/github.com/urfave/cli#DurationFlag

tonymbugua-outreach (approved):

cmd/devenv/provision/provision.go
kaldorn: While this might be true, that link is not accurate and v2 requires `EnvVars`.

PR #640, opened 2023-06-08
fix(kind): remove cgroup root config

kaldorn:
## Jira ID

[XX-XX]


jaredallard (approved):
Seems safe and not really required.

PR #641, opened 2023-06-10
feat(kubernetesruntime): add support for Loft

jaredallard:
Adds back support for Loft that was removed in https://github.com/getoutreach/devenv/pull/514.

Also upgrades Velero to better handle newer Kubernetes versions. Doesn't
use `central-hpm` (instead of `vcluster-fs-syncer`) because it doesn't
seem to function with Velero (mountPropagation issue?)

[DT-3802]

[DT-3802]: https://outreach-io.atlassian.net/browse/DT-3802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/embed/manifests/pre-restore/01_helmcontroller.jsonnet
```diff
@@ -1,57 +1,34 @@
 local cluster = import '../cluster.libsonnet';
 local ok = import '../libs.libsonnet';
+
```
jaredallard, line 3: Minor cleanup of this file to use our jsonnet library as well as it's own service account instead of mutating the default one :)

pkg/embed/manifests/pre-restore/metrics-server.jsonnet
jaredallard, line 1: Moved to `jsonnet` because ideally we should be using that everywhere here instead of `yaml`.

jkinkead:
Any way to split out Velero upgrade from the other work? We'd probably want to roll these out separately for testability / maintainability.

jaredallard:
> Any way to split out Velero upgrade from the other work? We'd probably want to roll these out separately for testability / maintainability.

I was getting issues without this on v1.23 GKE :/ I can try to detach it, but I have been getting extensive testing with both local and loft with this so far with no issues.

PR #642, opened 2023-06-12
feat: check if devenv is running before running commands

tonymbugua-outreach:
## Jira ID

DT-3267


cmd/devenv/apps/delete/delete.go
```diff
@@ -118,5 +119,9 @@ func (o *Options) Run(ctx context.Context) error {
 		return err
 	}
 
+	if kr.Status(ctx).Status.Status != status.Running {
```
jaredallard, line 122: Why not use `	devenvutil.EnsureDevenvRunning(ctx, conf, b)` instead? That does this logic and should be safer, since I believe `Status` could be nil here.

PR #645, opened 2023-06-14
feat(docker): add func `IsDockerRunning` to see if docker is running 

tonymbugua-outreach:
Provided an error message to users when docker is not running.

## Jira ID

DT-3242


kaldorn (changes requested):
This makes the comment on the `EnsureDevenvRunning` function is no longer accurate. I'm not 100% sure whether it's good hygiene or not, but I'd be inclined to have `IsDockerRunning` be renamed to `EnsureDockerIsRunning` and called prior to each call to `EnsureDevenvRunning`.

Alternatively, I'd update the `EnsureDevenvRunning` to be named more accurately and reflect what it actually is doing.

Edit: After chatting in person, I'm OK with the change as long as there's a comment explaining the function call here and updating the comment on the function -- i.e. If docker isn't running devenv can't be running, but this adds a more parseable/understandable error being returned

kaldorn (approved):

PR #647, opened 2023-06-15
feat(kubernetesruntime): add support for Loft

jaredallard:
Adds back support for Loft that was removed in https://github.com/getoutreach/devenv/pull/514.

Also upgrades Velero to better handle newer Kubernetes versions. Doesn't
use `central-hpm` (instead of `vcluster-fs-syncer`) because it doesn't
seem to function with Velero (mountPropagation issue?)

Disabled `prune` for loft because it relies on pruning the containerd image cache
which is KinD specific and doesn't apply to remote environments.

Put `tunnel` (localizer) behind a flag because it's a bad experience on
remote and given this is a "new" feature, we can disable that without
impacting the user experience much (`apps run` should be used instead.)

[DT-3802]

[DT-3802]: https://outreach-io.atlassian.net/browse/DT-3802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/devenv/provision/provision.go
```diff
@@ -414,7 +427,7 @@ func (o *Options) snapshotRestore(ctx context.Context) error { //nolint:funlen,g
 
 			for i := range pod.Spec.InitContainers {
 				cont := &pod.Spec.InitContainers[i]
-				if cont.Name == "restic-wait" {
+				if cont.Name == "restore-wait" {
```
jaredallard, line 430: velero upgrade changed the name from `restic-wait` to `restore-wait`

cmd/devenv/tunnel/tunnel.go
```diff
@@ -93,34 +104,27 @@ func (o *Options) Run(ctx context.Context) error {
...
-			case <-done:
-				return
-			case <-time.After(time.Second * 60):
-				// Sleep for 60 seconds between each alert after the first one.
-				continue
-			}
+	// Ensure that the user really wants to use this on a remote env, if we're
+	// running on a remote env. The experience is really poor, so we want to
+	// make sure that the user knows what they're doing.
+	if krun.GetConfig().Type == kubernetesruntime.RuntimeTypeRemote {
```
jaredallard, line 115: Just in case someone realllyyy wants to use it... but I could also be down to just not allow this.

pkg/embed/manifests/pre-restore/vault-secrets-operator.jsonnet
```diff
@@ -33,6 +33,9 @@ local manifests = {
         },
         limits: self.requests,
       },
+      serviceAccount: {
```
jaredallard, line 36: The chart operator somehow thinks service account secrets are generated automatically anymore... but they are. So we disable their manually created one.

internal/app/cmds/env.go
```diff
@@ -37,19 +36,16 @@ func (c *CmdWrapper) GetEnvVars(local bool, version, repositoryName string) ([]s
 		fmt.Sprintf("DEVENV_DEPLOY_DEV_IMAGE_REGISTRY=%s", DevImageRegistry),
 		fmt.Sprintf("DEVENV_DEPLOY_BOX_IMAGE_REGISTRY=%s", c.box.DeveloperEnvironmentConfig.ImageRegistry),
 		fmt.Sprintf("DEVENV_DEPLOY_APPNAME=%s", repositoryName),
-		fmt.Sprintf("DEVENV_TYPE=%s", kubernetesruntime.RuntimeTypeLocal),
+		fmt.Sprintf("DEVENV_TYPE=%s", c.rconf.Name),
+
+		// Keep around until applications have had a chance to update their
```
jaredallard, line 41: No longer used, but might still be required until https://github.com/getoutreach/stencil-golang/pull/308 lands everywhere, so we keep supplying it.

internal/app/cmds/devspace.go
```diff
@@ -35,26 +33,6 @@ func (c *CmdWrapper) EnsureDevspace() (string, error) {
 		return "", err
 	}
 
-	b, err := exec.Command(devspace, "list", "plugins").Output()
```
jaredallard, line 38: Removed these two functions as per [internal slack thread](https://outreach-hq.slack.com/archives/C03RV4VGAE7/p1686702439315139).

cmd/devenv/provision/prepare.go
```diff
@@ -119,20 +119,44 @@ func (o *Options) deployStage(ctx context.Context, stage string) error {
 		return err
 	}
 
-	// Deploy resourcer if we're a local runtime, we can only run things on a single node
-	// so we should mutate all pods to have zero resources.
-	// Special exeception is when we're generating snapshots.
-	if runtimeConf.Type == kubernetesruntime.RuntimeTypeLocal && os.Getenv("DEVENV_SNAPSHOT_GENERATION") == "" {
+	// Deploy resourcer
+	if os.Getenv("DEVENV_SNAPSHOT_GENERATION") == "" {
```
kaldorn, line 123: Does this skip deploying resourcer if we're generating snapshots? That seems fine. Is there resource contention around it?

internal/app/cmds/env.go
kaldorn: How is this no longer used?

kaldorn (approved):
Couple comments/questions. Nothing blocking.

Code is commented, things make sense.

cmd/devenv/provision/prepare.go
jaredallard: Yes, the deploying of resourcer is handled by devenv-snapshots which has a less ready PR here: https://github.com/getoutreach/devenv-snapshots/pull/116 (what I'm mainly still messing w/ right now)

internal/app/cmds/env.go
jaredallard: We no longer push an image into the cluster, which is what kind bin was for. Correct me if I'm wrong though (I did a search in the devspace yaml before removing it)

jaredallard: To clarify, this comment is about `DEVENV_KIND_BIN` and _not_ `DEVENV_TYPE`. Incorrectly selected that variable when I meant to select just the comment.

cmd/devenv/tunnel/tunnel.go
yj-yan: Just providing some perspective, I personally incline to not allow this if it is very high latency(unbearable high). It could return some support requests that we could not help in short term.

yj-yan (approved):

cmd/devenv/provision/prepare.go
```diff
@@ -119,20 +119,44 @@ func (o *Options) deployStage(ctx context.Context, stage string) error {
...
+				"kubectl", "rollout", "restart", "--namespace", "resourcer--bento1a", "deployment", "resourcer")
+			if err != nil {
+				return errors.Wrap(err, "failed to restart resourcer")
+			}
+
+			// wait for the rollout to complete
+			err = cmdutil.RunKubernetesCommand(ctx, "", false, os.Args[0], "--skip-update",
+				"kubectl", "rollout", "status", "--namespace", "resourcer--bento1a", "deployment", "resourcer")
+			if err != nil {
+				return errors.Wrap(err, "failed to restart resourcer")
```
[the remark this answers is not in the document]
jaredallard: Good point, will update!

cmd/devenv/tunnel/tunnel.go
jaredallard: I wanted to go ahead and remove this, but it turns out that some applications still require `devenv tunnel` for the reverse tunnel situation (e.g., giraffe). So, we have to keep this for now, but we can look into why they aren't using the Giraffe devspace or provide something _just_ for reverse tunnels (or use something off the shelf)

PR #657, opened 2023-06-22
fix: username in post restore template

marekfexa-outreach:
User user.Username (loginname) instead user.Name (with capital first letter and space) to fix this [issue](https://outreach-hq.slack.com/archives/C05CGT4CD9T/p1687433175252319)

## Jira ID

[XX-XX]


cmd/devenv/provision/post_restore.go
```diff
@@ -89,7 +89,7 @@ func (o *Options) getPostRestoreVariables(ctx context.Context) (*PostRestoreVari
 	}
 
 	vars := &PostRestoreVariables{
-		User:           u.Name,
+		User:           u.Username, // Username is login name (no spaces or capital characters)
```
jkinkead, line 92: It looks like this was changed from `u.Username` in [PR 630](https://github.com/getoutreach/devenv/pull/630/files#diff-cb173bc17bf43c116ed26ac359672e27cab418090dbdb6e0f40d26e9e99c662bL311), possibly accidentally.

cc: @jaredallard , in case this was deliberate.

I'm approving since this is a recent regression.

jkinkead (approved):

PR #669, opened 2023-07-01
feat(debug): add disk space problem detector

jaredallard:
Adds a `LowDiskSpace` problem detector for detecting when disk space is
low. Warning level is 80% while 90% is considered critical.

In order to implement this, I had to introduce a generic problem
detector. For now, it is any resource other than pod. Eventually we
should change this implementation to be more friendly for things outside
of Kubernetes (since this was largely designed as such for that). Left
comments in the code calling this out as I went.

No tests since this framework isn't very testable right now. Tested it
on Docker for Mac and orbstack to try to get a good idea of how it
handles different environments.

Wiki page is accessible here: https://github.com/getoutreach/devenv/wiki/LowDiskSpace


cmd/devenv/debug/debug.go
```diff
@@ -187,18 +193,18 @@ func (o *Options) Run(ctx context.Context) error { //nolint:funlen // Why: Best
...
-	if vaultProblems, is := o.getVaultWithProblems(ctx); is {
-		resourceProblems = append(resourceProblems, vaultProblems...)
+	if genericProblems, is := o.getGenericProblems(ctx); is {
+		resourceProblems = append(resourceProblems, genericProblems...)
 	}
 
 	bold.Println("done")
 	if len(resourceProblems) == 0 {
-		fmt.Println("Everything looks good 🎉")
+		fmt.Println("No problems detected ✅")
```
jaredallard, line 202: Edited this message to be less assuming, since "everything looks good" can be slightly annoying when a user knows there is an issue.

cmd/devenv/debug/debug.go
```diff
@@ -134,20 +137,21 @@ func (o *Options) getPodsWithProblems(ctx context.Context, pod *corev1.Pod) ([]R
 	return problems, len(problems) > 0
 }
 
-func (o *Options) getVaultWithProblems(ctx context.Context) ([]Resource, bool) {
+// getGenericProblems returns problems that aren't specific to Kubernetes
+// resources, e.g. Vault, disk space, or the like.
```
malept, line 141: Nit: "e.g." means "for example", so adding "or the like" is a bit redundant.
```suggestion
// resources, e.g. Vault or disk space.
```

cmd/devenv/debug/problem_os.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"os/exec"
+	"strconv"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// ProblemLowDiskSpace detects if the underlying Docker daemon is
+// running low on disk space, wether it's in the VM or on the host.
```
malept, line 22: Nit: typo
```suggestion
// running low on disk space, whether it's in the VM or on the host.
```

cmd/devenv/debug/problem_os.go
```diff
@@ -0,0 +1,84 @@
...
+// https://github.com/getoutreach/devenv/wiki/LowDiskSpace
+//
+// TODO(jaredallard): This only applies to local clusters, we don't have
+// the ability to selectively run problems yet.
```
malept, line 26: thought: consider passing in the runtime to the Detector function to determine whether this check needs to run.

cmd/devenv/debug/problem_os.go
```diff
@@ -0,0 +1,84 @@
...
+
+			out := scanner.Text()
+			fields := strings.Fields(out)
+			if len(fields) < 6 {
+				return "failed to parse df output: " + out, false, true
+			}
+
+			usedStr := fields[4]
+			if !strings.HasSuffix(usedStr, "%") {
+				return "failed to parse df output: " + fields[4], false, true
```
malept, line 65: Nit: distinguish between this error and the previous error
```suggestion
				return "failed to parse df Use% output: " + fields[4], false, true
```

malept (changes requested):
Since non-local clusters are a thing again, I think we should figure out how to not deal with this new check in that case.

Otherwise, this PR looks good.

PR #670, opened 2023-07-01
fix(loft): properly start and stop

jaredallard:
Fixes the commands for `start` and `stop` to support the newer Loft CLI
version. Stopped swallowing output for the respective commands when
invoked through `devenv start` and `devenv stop`. Aliased `sleep` and
`wakeup` to their respective devenv equivalents to better match the Loft
terminology.

Updated the pod readiness checker to better show `kubectl get pod`
information when a pod is pending (container statuses is sometimes
populated at that time).


deregtd (approved):
better wrapping of output on sleep/wakeup, with aliases.  also unified some of the pod status printout logic and added TODOs without ticket #s that somehow passed linting (prolly cuz this is open source and isn't running the advanced outreach-specific linters?)

PR #672, opened 2023-07-03
feat: Snapshot command to support target arg

Erik-Outreach:
To allow to split devenv-snapshots jobs to run one target per container, so that:
- Each target gets provision validated (currently only flagship is provision tested)
- Independent jobs helps isolating things (we can have base snapshot passing and flagship failing for example)
- Allows more snapshot targets avoiding the 1 hour limit of github credentials.

[QF-910]

[QF-910]: https://outreach-io.atlassian.net/browse/QF-910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/devenv/snapshot/snapshot.go
```diff
@@ -171,13 +171,16 @@ func NewCmdSnapshot(log logrus.FieldLogger) *cli.Command {
...
+					targetName := c.Args().First()
+
+					return o.Generate(c.Context, s, c.Bool("skip-upload"), box.SnapshotLockChannel(c.String("channel")), targetName)
 				},
 			},
 			{
 				Name:        "upload",
 				Description: "Upload all snapshots from local minIO container (outside local cluster)",
 				Usage:       "devenv snapshot upload",
+				ArgsUsage:   "targetName",
```
jkinkead, line 183:
```suggestion
				ArgsUsage:   "[targetName]",
```

cmd/devenv/snapshot/snapshot.go
```diff
@@ -254,7 +263,12 @@ func (o *Options) Generate(ctx context.Context, s *box.SnapshotGenerateConfig,
 		}
 	}
 
+	found := false
 	for name, t := range s.Targets {
+		if targetName != "" && targetName != name {
```
jkinkead, line 268: Add unit tests for new logic?

jkinkead (commented):
Style nits - and request for unit tests. :)

cmd/devenv/snapshot/snapshot.go
Erik-Outreach: I've updated existing test and created a test to validate args

Erik-Outreach: I think the existing `TestCanProvisionSnapshotDevenv` which was updated gives some coverage to the generate method. So I'll create new tests to cover `Upload` method, which have no test coverage at the moment.

jkinkead: "Some coverage", yes - but not separate tests of generating a specified target _and_ generating for ALL targets.

Erik-Outreach: I was thinking not to add test for all targets because I intend in a follow up PR to make the target parameter required for snapshot generation, deprecating the all targets option. Does it make sense?

Erik-Outreach: Fortunately this is only consumed in devenv-snapshots repo. So perhaps if we can we take a step back here, we could either:
- Support the two options (single target or all targets) for a short period of time. Once this version reaches stable channel and the change is rolled out in devenv-snapshots, the new attribute can be made optional.
OR, for simplification:
- Make the target required in this PR. On the same day, patch this to Stable and update devenv-snapshots accordingly. 

Would it be Okay to patch this change?

Erik-Outreach: Anyway I've updated this PR to cover the Upload function in tests. I propose I can simply remove the support for "all targets" as soon as it's not used by snapshot generation, which should be the next step.

jaredallard: Personally I think it's fine to not have a test for all targets given the context here.

jaredallard:
Disabling auto-merge because tests are failing. Holding off on approval until it's fixed. Please ping me for a re-review later!

cmd/devenv/snapshot/snapshot.go
Erik-Outreach: The change to cover "Upload" in the e2e tests was big enough for its own PR, so I've created it: https://github.com/getoutreach/devenv/pull/685. Once merged I can simply update it here passing the target.

Erik-Outreach:
> Disabling auto-merge because tests are failing. Holding off on approval until it's fixed. Please ping me for a re-review later!

@jaredallard ping :) Please review https://github.com/getoutreach/devenv/pull/685 and then this one 🙏

jaredallard:
> > Disabling auto-merge because tests are failing. Holding off on approval until it's fixed. Please ping me for a re-review later!
> 
> @jaredallard ping :) Please review #685 and then this one 🙏

Will wait for #685 to land fro any approvals, but so far it looks good to me.

jaredallard (approved):

tomassusovsky-outreach (approved):

PR #677, opened 2023-07-04
feat: Provision databases and base services first

Erik-Outreach:


cmd/devenv/provision/provision.go
```diff
@@ -313,31 +313,18 @@ func (o *Options) snapshotRestore(ctx context.Context) error { //nolint:funlen,g
 		return errors.Wrap(err, "failed to verify velero loaded snapshot")
 	}
 
-	o.log.Info("Starting snapshot restore (this may take awhile...)")
-	if err := m.Restore(ctx, snapshotTarget.VeleroBackupName); err != nil {
-		return errors.Wrap(err, "failed to restore snapshot")
-	}
+	stepOneNamespaces := []string{"bento1a", "kafka", "devenvtemporal--bento1a", "mint--bento1a"}
```
jkinkead, line 316: Can you document what these are and why they're in a separate step?

cmd/devenv/provision/provision.go
```diff
@@ -416,6 +403,32 @@ func (o *Options) snapshotRestore(ctx context.Context) error { //nolint:funlen,g
 		return ctx.Err()
 	}
 
+	// wait for stateful sets
+	o.log.Info("Waiting for statefulsets to be ready...")
```
jkinkead, line 407: It seems like you're waiting for a set of namespaces, and logging that would be clearer.

internal/snapshot/from_snapshot.go
```diff
@@ -112,16 +112,19 @@ func (m *Manager) deleteExistingRestore(ctx context.Context, snapshotName string
 }
 
 // Restore restores a given snapshot into an environment.
-func (m *Manager) Restore(ctx context.Context, snapshotName string) error { //nolint:funlen // Why: there are no reusable parts to extract
+func (m *Manager) Restore(ctx context.Context, snapshotName string, skipRetrieve bool,
```
jkinkead, line 115: Document new function parameters.

internal/snapshot/from_snapshot.go
```diff
@@ -112,16 +112,19 @@ func (m *Manager) deleteExistingRestore(ctx context.Context, snapshotName string
...
+	includedNamespaces, excludedNamespaces []string) error { //nolint:funlen // Why: there are no reusable parts to extract
```
jkinkead, line 116: I don't like this linter rule, but if you pulled out some of the more complicated logic into unit-testable code, I bet this warning would go away.

jkinkead (commented):
I would like to see unit tests.

Per comments in Slack, doing this in a slightly more generic way (like having a modeled "step" struct with the list of namespaces, and a list of these steps for a restore) would give us a few more tools to use in ordering startups and making this less brittle.

cmd/devenv/provision/provision.go
```diff
@@ -313,9 +342,46 @@ func (o *Options) snapshotRestore(ctx context.Context) error { //nolint:funlen,g
...
+		"vault-secrets-operator",
+		"local-path-storage",
+		"monitoring",
+		"resourcer--bento1a"}
+
+	// for testing, inject restore steps
+	//stageZeroNamespaces := []string{"bento1a", "kafka", "devenvtemporal--bento1a", "mint--bento1a"}
+	// snapshotTarget.RestoreSteps = []RestoreStep{
+	// 	{IncludedNamespaces: stageZeroNamespaces},
+	// 	{ExcludedNamespaces: append(stageZeroNamespaces, defaultExcludedNamespaces...)},
```
Erik-Outreach, line 362: 💡 we can probably remove the `defaultExcludedNamespaces` and let ExcludedNamespaces be defined in snapshot generation config

cmd/devenv/provision/provision.go
Erik-Outreach: in my most recent change I'm proposing to move it to devenv-snapshots repo, in this case so the docs would go there

cmd/devenv/provision/provision.go
```diff
@@ -259,6 +259,59 @@ func NewCmdProvision(log logrus.FieldLogger) *cli.Command {
...
+
+	// PostDeployApps is an array of applications to deploy via deploy-app
+	// after running the Command specified.
+	PostDeployApps []string `yaml:"post_deploy_apps"`
+
+	// ReadyAddress is a URL to ping before marking the devenv as ready
+	ReadyAddress string `yaml:"readyAddress"`
+
+	// RestoreSteps defines which steps are used for restore
+	RestoreSteps []RestoreStep `yaml:"restore_steps"`
```
Erik-Outreach, line 303: This is the proposed property. If this idea is good, next step I'll add this prop to gobox project, then get this PR ready.

cmd/devenv/provision/provision.go
Erik-Outreach: I'm proposing that here: https://github.com/getoutreach/devenv-snapshots/pull/130/files

cmd/devenv/provision/provision.go
Erik-Outreach: Here's the PR to move this from here: https://github.com/getoutreach/gobox/pull/324/files

internal/e2e/e2e_test.go
```diff
@@ -102,6 +102,35 @@ func TestSnapshotArgIsValid(t *testing.T) {
 }
 
 func TestCanProvisionSnapshotDevenv(t *testing.T) {
+	config := &box.SnapshotGenerateConfig{
```
Erik-Outreach, line 105: This test is a bit of an overkill, and it takes a long time to run. I'm thinking to remove it, now or once main flagship snapshot have steps defined: https://github.com/getoutreach/devenv-snapshots/pull/130. Ideally we shouldn't need a whole e2e test to cover each new provision scenario

cmd/devenv/provision/provision.go
```diff
@@ -316,9 +316,57 @@ func (o *Options) snapshotRestore(ctx context.Context) error { //nolint:funlen,g
...
 	}
 
-	o.log.Info("Starting snapshot restore (this may take awhile...)")
-	if err := m.Restore(ctx, snapshotTarget.VeleroBackupName); err != nil {
-		return errors.Wrap(err, "failed to restore snapshot")
+	if len(snapshotTarget.Config.RestoreSteps) == 0 {
+		o.log.Info("No restore steps defined for this snapshot. Using the default configuration.")
+
+		// TODO[DTSS-829]: This should be moved into the generation framework
+		defaultExcludedNamespaces := []string{
```
jaredallard, line 323: These namespaces need to be ignored for the provision to succeed, so I think that maybe we should always skip these? Maybe always inject this in as being excluded if excluded is passed?

All the options seem heavy to implement, but I worry that not having these be removed by default will lead to a bug in the future.

Erik-Outreach: I'm proposing to move these to devenv-snapshots job ([link](https://github.com/getoutreach/devenv-snapshots/pull/130/files#diff-ef1fe88c7166dd337ae5aaff6da3773a54bbde73c74223f7fa61cef38984def1R27)) which seems to be a better place for such exclusions, making devenv less aware of details of the snapshot. Obviously if we forget to add the exclusion rules there, it won't work, but we're safe guarded by PR validation going through provision to confirm it.

jaredallard: @Erik-Outreach That would make sense if this wasn't a hard requirement for _all_ snapshots to be able to be restored. Those namespaces must be excluded because they are installed via helm charts before restoring the snapshot. If restored on top, they could be for the wrong Kubernetes runtime or have other hard-to-debug issues, so they were hardcoded into the devenv as being excluded.

I don't think relying on all snapshots to have the same duplicated excluded_namespaces field is very sustainable, especially since you're going ahead with the per-team snapshots, which would lead to us having more and more of them.

Erik-Outreach: > Those namespaces must be excluded because they are installed via helm charts before restoring the snapshot.

Makes sense, so those probably belong more in devenv codebase than elsewhere. Agree to make this exclusion rule apply in all cases

PR #685, opened 2023-07-12
test: Cover "Upload" in e2e test

Erik-Outreach:
- For covering Upload in tests, obviously.
- Also to unblock [this PR ](https://github.com/getoutreach/devenv/pull/672) where Upload method was changed, and test coverage was asked.

[QF-910]

[QF-910]: https://outreach-io.atlassian.net/browse/QF-910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/snapshotstaging/snapshotstaging.go
```diff
@@ -57,7 +57,12 @@ func NewSnapShotOptions(ctx context.Context, log logrus.FieldLogger, b *box.Conf
 }
 
 // Provisions a minIO instance outside devenv cluster to store snapshots for validation
```
jaredallard, line 59: Update the comment?

pkg/snapshotstaging/snapshotstaging.go
```diff
@@ -57,7 +57,12 @@ func NewSnapShotOptions(ctx context.Context, log logrus.FieldLogger, b *box.Conf
...
-func (o *SnapshotOptions) ProvisionStagingStorage() error {
+func (o *SnapshotOptions) ProvisionDefaultStagingStorage() error {
+	return o.ProvisionStagingStorage(defaultMinioContainerName, PortNumber)
+}
+
+// Provisions a minIO instance outside devenv cluster to store snapshots for validation
```
jaredallard, line 64: Update the comment here as well?

pkg/snapshotstaging/snapshotstaging.go
```diff
@@ -236,8 +241,22 @@ func (o *SnapshotOptions) GetMinioIPEndpointFromInsideCluster() (string, error)
...
+	os.Setenv("AWS_ACCESS_KEY_ID", AccessKey)
+	os.Setenv("AWS_SECRET_ACCESS_KEY", SecretKey)
+}
+
+// RedirectS3ToMinioForProvision redirects to s3 to default minio for provision, from inside the cluster
+func (o *SnapshotOptions) RedirectS3ToDefaultMinioForProvision() error {
+	return o.RedirectS3ToMinioForProvision(defaultMinioContainerName)
+}
+
+// RedirectS3ToMinioForProvision redirects to s3 to specified minio for provision, from inside the cluster
```
jaredallard, line 258: This comment isn't entirely clear how it redirects to s3. I'd say that it affects the box configuration and changes the SnapshotConfig to point to the specified minio container, otherwise it feels a little magical. Also, in general, we should document side effects in comments.

pkg/snapshotstaging/snapshotstaging.go
```diff
@@ -236,8 +241,22 @@ func (o *SnapshotOptions) GetMinioIPEndpointFromInsideCluster() (string, error)
 	return ipAddr, nil
 }
 
-// Redirects s3 to minio when running from inside the cluster
-func (o *SnapshotOptions) RedirectS3ToMinioFromCluster() error {
+// RedirectS3ToMinioFromOutsideCluster redirects to s3 for upload method, only used in tests.
```
jaredallard, line 244: Same goes here.

pkg/snapshotstaging/snapshotstaging.go
```diff
@@ -266,20 +285,26 @@ func (o *SnapshotOptions) RedirectS3ToMinioFromCluster() error {
 	return nil
 }
 
+// Destroys default minIO container
```
jaredallard, line 288: Update comment to reflect new function name?

pkg/snapshotstaging/snapshotstaging.go
```diff
@@ -266,20 +285,26 @@ func (o *SnapshotOptions) RedirectS3ToMinioFromCluster() error {
...
+func (o *SnapshotOptions) CleanUpDefault() error {
+	return o.CleanUp(defaultMinioContainerName)
+}
+
 // Destroys minIO container
```
jaredallard, line 293:
```suggestion
// CleanUp removes the provided minio container
```

pkg/snapshotstaging/snapshotstaging.go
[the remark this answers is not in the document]
Erik-Outreach: Thanks for the idea, I like it, to the point I had to it now :) So I've created the minio struct. Couldn't use a singleton because I added a context object to the struct definition. Let me know what you think

pkg/snapshotstaging/minio.go
```diff
@@ -0,0 +1,205 @@
...
+	containerName string
+	hostPort      string
+	log           logrus.FieldLogger
+	b             *box.Config
+	ctx           context.Context
+}
+
+// NewMinIO creates a minIO instance by container and port.
+// It allows multiple minIO instances side by side as long as they have different container name and port
+func NewMinIO(ctx context.Context, containerName, hostPort string, log logrus.FieldLogger, b *box.Config) *MinIOInstance {
```
jaredallard, line 42: Generally, don't embed `ctx` into a struct. Instead, take it as an option in a function that consumes it.

The logic behind this is that `ctx` is highly dependent on when something is calling a function, so the client could be created in a completely different environment than the function that is being called at the time. This would make it difficult to control when `Provision` is canceled, for example, if you were in a different context when creating the instance.

For `box.Config`, I would also suggest passing that to `UseSnapshotStorage` instead since it'll be easier to add tests for.

In general, I'd like to see tests for this! A lot of the motivation for making this its own struct was to be able to be tested better.

pkg/snapshotstaging/minio.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+const minioImage = "minio/minio:RELEASE.2022-02-05T04-40-59Z"
+
+const AccessKey = "ACCESS_KEY"
+const SecretKey = "SECRET_KEY"
+const BucketName = "outreach-devenv-snapshots"
+const HostIP = "127.0.0.1"
+
+type MinIOInstance struct {
```
jaredallard, line 32: Exported type, so please add a comment as a header in the format of `// MinIOInstance ...`

pkg/snapshotstaging/snapshotstaging.go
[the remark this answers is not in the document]
jaredallard: Left some more comments, this is great progress though! Really excited for it 🎉

pkg/snapshotstaging/minio.go
Erik-Outreach: That makes sense - removed ctx and box.config from the struct for testability and better context handling.

pkg/snapshotstaging/snapshotstaging.go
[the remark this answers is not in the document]
Erik-Outreach: Tks for the reviews, I think it's ready for another round now :)

pkg/snapshotstaging/minio.go
jaredallard, line 1: Implementation looks good, it'd be nice to add some tests _just_ for the minio struct here. It should be pretty simple thanks to the break out!

jaredallard: And, yes, I do see the "e2e" test above. That's good enough for this PR to land, IMO, but noting this for the future.

jaredallard (approved):
LGTM, with comment about having dedicated tests for the new minio struct being nice to have, but not required because we get sufficient coverage with the `e2e` test right now.

Erik-Outreach:
> LGTM, with comment about having dedicated tests for the new minio struct being nice to have, but not required because we get sufficient coverage with the `e2e` test right now.

Agree, I'm not sure how to add the minio struct test right now, but I'm sure future me will handle it nicely :D

PR #687, opened 2023-07-14
feat(context): re-enable showing all available Loft clusters

jaredallard:
This re-enables the ability to see all Loft clusters when running
`devenv context`. Previously, you used to need to pull a kube config for
every cluster that was shown in the `devenv context` list. Because we
need to reach out of Loft each time, this was slow. As such, when Loft
started to be used more, we short-circuited that function to only show
the current user's devenv. With #671 landing certificate sharing, it's
time to support shared environments again.

This PR mitigates the past performance issues by making `GetKubeConfig`
take the name of the cluster that it should return a kubeconfig for
(with KinD always being the default one). As such, we no longer need to
return the kubeconfig in the `RuntimeCluster` type (used for listing).
With this change, I change all consumers of that field to now use
`GetKubeConfig()` with the name of the context in question. For
`provision` this was as simple as passing the default cluster name in,
while for `context` this was just the new context name.

Also cleaned up some documentation as I went.

[DT-3860]

[DT-3860]: https://outreach-io.atlassian.net/browse/DT-3860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/devenv/destroy/destroy.go
```diff
@@ -46,7 +46,7 @@ type Options struct {
 }
 
 func NewOptions(log logrus.FieldLogger, b *box.Config) (*Options, error) {
-	d, err := dockerclient.NewClientWithOpts(dockerclient.FromEnv)
+	d, err := dockerclient.NewClientWithOpts(dockerclient.FromEnv, dockerclient.WithAPIVersionNegotiation())
```
jaredallard, line 49: Forgot to mention in the PR description. I updated the Docker dependency to try to fix an issue I was running into (wasn't it), as part of that I moved us to always use `WithAPIVersionNegotiation` to make sure future docker upgrades in the `go.mod` never break anything. So, all of these changes are just making that safer (we used it sporadically beforehand).

There's also a `timeout` change later on that moves to using `container.StopOptions`, a much better way of specifying options to stop containers than before.

cmd/devenv/provision/provision.go
```diff
@@ -752,13 +753,13 @@ func (o *Options) Run(ctx context.Context) error { //nolint:funlen,gocyclo // Wh
 		return errors.Wrap(err, "failed to setup image pull secret")
 	}
 
-	if err := o.generateDockerConfig(); err != nil { //nolint:govet // Why: OK w/ err shadow
```
jaredallard, line 755: Removed the `govet` no lints here because they no longer apply.

internal/kubernetesruntime/kind.go
```diff
@@ -95,6 +95,7 @@ func (kr *KindRuntime) Status(ctx context.Context) RuntimeStatus {
 	if err != nil {
 		if dockerclient.IsErrNotFound(err) {
 			resp.Status.Status = status.Unprovisioned
+			resp.Status.Reason = fmt.Sprintf("container %s not found: %v", containerruntime.ContainerName, err)
```
jaredallard, line 98: Added some better debug information for when this does happen.

cmd/devenv/context/context.go
```diff
@@ -27,20 +27,24 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+// Options contains the arguments and required dependencies of the
+// context command.
 type Options struct {
 	log            logrus.FieldLogger
+	b              *box.Config
 	DesiredContext string
```
malept, line 35: thought: I'm kind of surprised `Options` needed docs, but `DesiredContext` (a public property) did not.

malept (approved):

cmd/devenv/context/context.go
jaredallard: Oof, don't know how I missed that... I'll fix that.

PR #692, opened 2023-07-14
fix: Avoid deleting pods after first deploy

Erik-Outreach:


yj-yan (approved):
LGTM

internal/app/deploy.go
```diff
@@ -122,7 +122,12 @@ func (a *App) deployBootstrap(ctx context.Context) error {
 			if err := a.buildDockerImage(ctx); err != nil {
 				return errors.Wrap(err, "failed to build image")
 			}
-			builtDockerImage = true
+
+			// check if namespace was deployed
+			err := cmdutil.RunKubernetesCommand(ctx, a.Path, false, "get", "namespace", fmt.Sprintf("%s--bento1a", a.RepositoryName))
```
jaredallard, line 127: Instead of assuming the repository namespace and relying on `kubectl`, why not use the apps client to check if the application has been deployed before?

jaredallard: This won't work. `get` is being parsed as the command. 

Try doing something like this instead:

```go
cmdutil.RunKubernetesCommand(ctx, "", false, os.Args[0], "--skip-update", "kubectl", "get", "namespace" <write the rest>)
```

Erik-Outreach: Good idea about using apps client to check it. Looking at this PR now, I probably made a screw up with my changes because this is not the latest version I remember testing 🤔 . Anyway thanks for following up and fixing my mess

PR #693, opened 2023-07-14
fix(dfm): adjust more settings

jaredallard:
Manages more Docker for Mac settings:

* `dockerBinInstallPath` - If not `system`, some docker commands will
  fail (namely `docker build`).
* `enableDefaultDockerSocket` - If not `true`, KinD's docker runtime
  will fail to detect the cluster sometimes. Unclear why, but also a lot
  of things depend on `/var/run/docker.sock` in the wild.
* `requireVmnetd` - Without this, we can't bind to privileged ports.
  Special case we need to ensure `vmnetd` is installed, because Docker
  For Mac doesn't do it. We do so here.

Adjusted the Docker for Mac start logic to always try to open the app.
This handles then Docker for Mac daemon sometimes (more often lately?)
crashing when we tell it to quit, which `open -a Docker.app` doesn't
fail when that happens, so a user might wait forever. Doing it on each
check doesn't open a window or seem to do anything other than start it
again, if it's not running, so this works around that nicely.

A user will still need to click "quit" if Docker for Mac crashes, but
this seems like the best we can do.


cmd/devenv/provision/docker_for_mac.go
```diff
@@ -17,17 +17,20 @@ import (
...
 	// Give Docker for Mac time to stop.
 	time.Sleep(2 * time.Second)
 
-	cmd := exec.CommandContext(ctx, "open", "-a", "Docker")
-	if out, err := cmd.CombinedOutput(); err != nil {
-		return errors.Wrapf(errors.Wrap(err, string(out)), "failed to open Docker for Mac (try starting it manually)")
-	}
-
-	ticker := time.NewTicker(7 * time.Second)
+	ticker := time.NewTicker(5 * time.Second)
```
jaredallard, line 25: 7 feels weird, lowered to 5.

cmd/devenv/provision/docker_for_mac.go
```diff
@@ -74,20 +79,44 @@ func reconcileDockerForMacConfig(_ context.Context, settingsFile string) (bool,
 		}
 	}
 
-	// This enables support for older containers on macOS. This can be
```
jaredallard, line 77: Didn't call this out, we don't need cgroupv1 support anymore.

jaredallard: Maybe we should set this to `false`? Open thought.

cmd/devenv/provision/docker_for_mac.go
malept: Maybe it needed to be 7 on Intel Macs? Just guessing.

cmd/devenv/provision/docker_for_mac.go
```diff
@@ -17,17 +17,20 @@ import (
...
-	}
-
-	ticker := time.NewTicker(7 * time.Second)
+	ticker := time.NewTicker(5 * time.Second)
 	for {
+		// Always try to start Docker for Mac, even if it's already running.
+		// This doesn't appear to have any side effects.
+		cmd := exec.CommandContext(ctx, "open", "-a", "Docker")
+		if out, err := cmd.CombinedOutput(); err != nil {
+			return errors.Wrapf(errors.Wrap(err, string(out)), "failed to open Docker for Mac (try starting it manually)")
```
malept, line 31: Nit: unsure why this is `errors.Wrapf` and not just `errors.Wrap`.
```suggestion
			return errors.Wrap(errors.Wrap(err, string(out)), "failed to open Docker for Mac (try starting it manually)")
```

cmd/devenv/provision/docker_for_mac.go
malept: Yeah, I'd probably just set it to `false`.

cmd/devenv/provision/docker_for_mac_test.go
```diff
@@ -30,6 +34,33 @@ func TestDockerConfigRead(t *testing.T) {
...
+		{
+			Name: "EnsureDockerBinInstallPathSystem",
+			Given: dockerForMacSettings{
+				DockerBinInstallPath: "user",
+			},
+			Accepted: dockerForMacSettings{
+				DockerBinInstallPath: "system",
+			},
+		},
+		{
```
jaredallard, line 55: Didn't add a test for the vmnetd behaviour because we don't have the ability to test that in CI right now.

cmd/devenv/provision/docker_for_mac.go
```diff
@@ -43,15 +46,22 @@ func startDockerForMac(ctx context.Context, d dockerclient.APIClient, log logrus
...
-// It's referenced here to fix some graphical problems based on specific configuration here: https://docs.docker.com/desktop/faqs/general/
-func reconcileDockerForMacConfig(_ context.Context, settingsFile string) (bool, error) { //nolint:funlen,lll // Why: there are no reusable parts to extract
+// reconcileDockerForMacConfig ensures that Docker for Mac is configured
+// correctly for use with devenv.
+//
+// This works with the largely undocumented `settings.json` file that
+// Docker For Mac uses to store user configuration. This file lives at
+// ~/Library/Group Containers/group.com.docker/settings.json
+//
+//nolint:funlen,lll,gocyclo // Why: there are no reusable parts to extract
```
malept, line 56: Nit: it'd be nice to annotate why gocyclo was added to the nolint list.

malept (approved):

cmd/devenv/provision/docker_for_mac.go
jaredallard: If I didn't have to wait for tests I would. The comment there still makes sense, to me.

PR #703, opened 2023-07-19
feat: check devenv status before running commands

jaredallard:
This PR changes most commands to assert that the devenv is in a valid
state before running. Most commands need it to be in the `running`
state, so `EnsureDevenvIsRunning` is changed to do that (as the name
implies).

Other commands have been changed (stop/start) to assert that the devenv
is in a provisioned state instead.

Also updated the devcontainer to work out of the box with asdf.

[DT-3267]

[DT-3267]: https://outreach-io.atlassian.net/browse/DT-3267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/devenv/snapshot/snapshot.go
```diff
@@ -120,9 +120,11 @@ func NewOptions(log logrus.FieldLogger, b *box.Config) (*Options, error) {
...
 func NewCmdSnapshot(log logrus.FieldLogger) *cli.Command {
 	return &cli.Command{
-		Name:        "snapshot",
-		Usage:       "Manage snapshots of your developer environment",
-		ArgsUsage:   "[targetName]",
+		Name:      "snapshot",
+		Usage:     "Manage snapshots of your developer environment",
+		ArgsUsage: "[targetName]",
+		// We hide this command because it's not meant for end users to use.
+		Hidden:      true,
```
jaredallard, line 127: Hid the `snapshots` command since it's meant just for snapshot generation right now.

internal/devenvutil/devenv.go
```diff
@@ -41,23 +41,31 @@ import (
...
+		return nil, fmt.Errorf("developer environment is not running, please run 'devenv start'")
+	}
+
 	return r, nil
 }
 
 // EnsureDockerIsRunning checks if docker is running. Returns a boolean
 func EnsureDockerIsRunning() (bool, error) {
-	cmd := exec.Command("docker", "info")
+	cmd := exec.Command("docker", "version")
```
jaredallard, line 68: Changed to `docker version` because its 2x as fast.

kaldorn (approved):
That cleans things up quite a bit. Good work.

PR #705, opened 2023-07-20
feat(kind): support 350 pods

jaredallard:
Updates the maxPods to 350. Being limited by `maxPods` doesn't really help us discover any perf issues, so keeping it low doesn't make a ton of sense.

We've also surpassed 250 pods in some devenvs, so this gives us a lot of room.


deregtd (approved):
maxPods: 250 -> 350

